PR #5, opened 2015-12-01
Add and use methods to PromiseState for parity with Promise

ryanbrainard:
I added the remaining methods to `PromiseState` for parity with `Promise`. I did this because I needed to error handle a `PromiseState` that was part of a composition, and this made it clean use with `catch()` as it were an actual `Promise`.

For example, sometimes `statusesFetch` will fail, but this recovers the error and sets the value to `[]` and makes it still work with `PromiseState.all()` (I would not have been able to recover inside `AlarmTable` because the composition would have never been fulfilled):

```
render() {
    const { alarmsFetch, statusesFetch } = this.props

    return (
      <PromiseStateContainer
        ps={PromiseState.all([alarmsFetch, statusesFetch.catch(() => [])])}
        onFulfillment={([alarms, statuses]) => <AlarmTable alarms={alarms} statuses={statuses}/>}
      />
    )
  }
``` 

cc: @jsullivan @ricardochimal

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -63,6 +63,21 @@ Properties should be treated as read-only and immutable. If the `Promise` enters
 
 ##### Static Methods
 
+##### `create(meta): PromiseState`
+
+##### `refresh(previous: PromiseState, meta): PromiseState`
+
+##### `resolve(value, meta): PromiseState`
+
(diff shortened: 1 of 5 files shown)


ryanbrainard:
Merged locally. Closing.

PR #6, opened 2015-12-03
Chainable requests

ryanbrainard:
This introduces four new attributes for chainable requests. I'm still not sure on the names and syntax, but here they are:

- `then(value, meta): request` *(Function)*: returns a request to fetch after fulfillment of this request and replaces this request. Takes the `value` and `meta` of this request as arguments.
- `catch(reason, meta): request` *(Function)*: returns a request to fetch after rejection of this request and replaces this request. Takes the `value` and `meta` of this request as arguments.
- `andThen(value, meta): { prop: request, ... }` *(Function)*: returns an object of request mappings to fetch after fulfillment of this request but does not replace this request. Takes the `value` and `meta` of this request as arguments.
- `andCatch(reason, meta): { prop: request, ... }` *(Function)*: returns an object of request mappings to fetch after rejection of this request but does not replace this request. Takes the `value` and `meta` of this request as arguments.

`then` is helpful for cases where multiple requests are needed to get to the data needed by the component and the subsequent requests rely on data from the previous requests. For example, if you need to make a request to look up a piece of data to perform some other operation:
 
      connect(({ name }) => ({
        barFetch: {
          url: `/foos/${name}`,
          then: (foo) => `/bar-for-foos-by-id/${foo.id}`
        }
      }))

`andThen` is similar, but is intended for side effect requests where you still need access to the result of the first request and/or need to fanout to multiple requests:

      connect(({ name }) => ({
        fooFetch: {
          url: `/foos/${name}`,
          andThen: (foo) => { 
            barFetch: `/bar-for-foos-by-id/${foo.id}` 
          }
        }
      }))

This is also helpful for cases where a fetch function is changing data that is in some other fetch that is a collection. For example, if you have a list of `foo`s and you create a new `foo`, the list needs to be refreshed:

     connect(({ name }) => ({
        foosFetch: '/foos',
        createFoo: (name) => {
           method: 'POST',
           url: '/foos',
           andThen: () => { 
             foosFetch: { 
               url: '/foos', 
               refreshing: true 
             }
           }
        }
      }))

`catch` and `andCatch` are similar, but for error cases.

Feedback is welcome on the naming of these attributes. Are they clear what they do and how they work?

cc: @jsullivan @ricardochimal

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -174,8 +174,6 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
               }, mapping.refreshInterval)
             }
 
-            // TODO: re-think naming and params of then, catch, andThen, andCatch
-
             if (Function.prototype.isPrototypeOf(mapping.then)) {
               this.refetchDatum(prop, coerceMapping(null, mapping.then(value, meta)))
               return
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -174,15 +174,27 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
               }, mapping.refreshInterval)
             }
 
+            // TODO: re-think naming and params of then, catch, andThen, andCatch
+
+            if (Function.prototype.isPrototypeOf(mapping.then)) {
+              this.refetchDatum(prop, coerceMapping(null, mapping.then(value, meta)))
+              return
+            }
+
             this.setAtomicState(prop, startedAt, mapping, PromiseState.resolve(value, meta), refreshTimeout, () => {
-              if (Function.prototype.isPrototypeOf(mapping.then)) {
-                this.refetchDataFromMappings(mapping.then(value, meta))
+              if (Function.prototype.isPrototypeOf(mapping.andThen)) {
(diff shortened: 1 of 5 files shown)


src/components/connect.js
```diff
@@ -174,14 +174,34 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
               }, mapping.refreshInterval)
             }
 
-            this.setAtomicState(prop, startedAt, mapping, PromiseState.resolve(value, meta), refreshTimeout)
+            // TODO: re-think naming and params of then, catch, andThen, andCatch
```
jsullivan: The naming of these makes sense to me, especially given the context that @ryanbrainard says inspired these functions:

http://www.scala-lang.org/files/archive/nightly/docs/library/index.html#scala.concurrent.Future

jsullivan:
I like all of this. Thumbs up from me.

ryanbrainard:
ok, i'm going to pull out that re-naming TODO, bump the version, merge and publish this

jsullivan:
Sounds good. After you a new release is cut then I'm good with https://github.com/heroku/dogwood-ops/pull/84 getting merged in, too.

PR #7, opened 2015-12-17
Add javascript syntax color

BafS:
More pleasant and easy to read


ryanbrainard:
Thanks, but went with #8 since it uses JSX.

PR #12, opened 2015-12-18
fix typo in README

kmoe:


ryanbrainard:
Sorry to be a pain, but the way GitHub is showing this diff with the soft wraps, I can't tell what changed. Can you add a quick PR description?

kmoe:
Just a s/let's/lets in "instead of mapping state to props, this let's you map props to URLs to props" -> "instead of mapping state to props, this lets you map props to URLs to props"

kmoe:
Ah, didn't noticed you'd merged a subsequent PR with this fix.

ryanbrainard:
Thanks anyway! :)

PR #23, opened 2015-12-26
Make it possible to disable auto refresh while browser tab is hidden

hekike:
- [x] provide `refreshHidden` option to disable bg refresh while tab is hidden

Pooling the server while tab is hidden is usually wrong: unnecessary load on servers, bandwitch usage and battery draining on the client.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,10 +51,12 @@
     "react-addons-test-utils": "^0.14.0",
     "react-dom": "^0.14.0",
     "rimraf": "^2.3.4",
+    "sinon": "^1.17.2",
     "webpack": "^1.11.0"
   },
   "dependencies": {
     "hoist-non-react-statics": "^1.0.3",
+    "visibilityjs": "^1.2.1",
     "invariant": "^2.0.0"
   },
   "peerDependencies": {
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -6,6 +6,7 @@ import shallowEqual from '../utils/shallowEqual'
 import PromiseState from '../PromiseState'
 import hoistStatics from 'hoist-non-react-statics'
 import invariant from 'invariant'
+import visibility from 'visibilityjs'
 
 const defaultMapPropsToRequestsToProps = () => ({})
 
@@ -48,6 +49,8 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
 
     invariant(isPlainObject(mapping), 'Request for `%s` must be either a string or a plain object. Instead received %s', prop, mapping)
     invariant(mapping.url, 'Request object for `%s` must have `url` attribute.', prop)
+
+    mapping.refreshHidden = mapping.refreshHidden === undefined ? true : mapping.refreshHidden
 
     mapping.equals = function (that) {
       if (this.comparison !== undefined) {
@@ -153,9 +156,21 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
 
       refetchDatum(prop, mapping) {
         const startedAt = new Date()
+        const hasRefreshInterval = mapping.refreshInterval > 0
 
         if (this.state.refreshTimeouts[prop]) {
           window.clearTimeout(this.state.refreshTimeouts[prop])
+        }
+
+        // do not refresh while tab is hidden and refreshHidden is unset
+        if (hasRefreshInterval && !mapping.refreshHidden && visibility.hidden()) {
+          const visibilityListener = visibility.change(() => {
+            if (!visibility.hidden()) {
+              visibility.unbind(visibilityListener)
+              this.refetchDatum(prop, mapping)
+            }
+          })
+          return
         }
 
         const request = buildRequest(mapping)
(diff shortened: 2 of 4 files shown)


ryanbrainard:
I probably won't be able to look into this in-depth until next week, but just noticed that `visibility` is a devDependency. Is that correct? Also, the defaulting of request attribute should all be in `buildRequest()` if possible.

ryanbrainard:
cc: @ricardochimal any thoughts on this since you originally had the use case for refresh intervals?

src/components/connect.js
```diff
@@ -153,11 +156,23 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
...
+        const isRefresh = mapping.refreshInterval > 0
 
         if (this.state.refreshTimeouts[prop]) {
           window.clearTimeout(this.state.refreshTimeouts[prop])
         }
 
+        // do not refresh while tab is hidden and refreshHidden is unset
+        if (isRefresh && !mapping.refreshHidden && visibility.hidden()) {
+          const visibilityListener = visibility.change(() => {
+            if (!visibility.hidden()) {
```
ryanbrainard: Looks like this could be simplified with [`visibility. onVisible()`](https://github.com/ai/visibilityjs#waiting-until-the-page-becomes-visible) to improve readability and avoid the `if (!visibility.hidden()) {` conditional.

src/components/connect.js
```diff
@@ -153,11 +156,23 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
...
       refetchDatum(prop, mapping) {
         const startedAt = new Date()
+        const isRefresh = mapping.refreshInterval > 0
 
         if (this.state.refreshTimeouts[prop]) {
           window.clearTimeout(this.state.refreshTimeouts[prop])
         }
 
+        // do not refresh while tab is hidden and refreshHidden is unset
+        if (isRefresh && !mapping.refreshHidden && visibility.hidden()) {
```
ryanbrainard: It's a corner case, but it would probably be good to also check for `this.state.mappings[prop] && this.state.mappings[prop].fulfilled` so that this does not block the initial load in case the tab is hidden.

src/components/connect.js
```diff
@@ -49,6 +50,8 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
     invariant(isPlainObject(mapping), 'Request for `%s` must be either a string or a plain object. Instead received %s', prop, mapping)
     invariant(mapping.url, 'Request object for `%s` must have `url` attribute.', prop)
 
+    mapping.refreshHidden = mapping.refreshHidden === undefined ? true : mapping.refreshHidden
```
ryanbrainard: Ignore what I said previously about moving this into `buildRequest()`. For #31, I need to break out the defaults from the request building. Once that is done, this can move there (no need to block on that in the meantime though).

src/components/connect.js
```diff
@@ -153,11 +156,23 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
 
       refetchDatum(prop, mapping) {
         const startedAt = new Date()
+        const isRefresh = mapping.refreshInterval > 0
```
ryanbrainard: not sure i like the name of this variable. i'm not really sure what to call it, but this would still be true on initial page load and could also get confusing with the `refreshing` attribute. Perhaps the clearest would be a more literal `hasRefreshInterval`

test/components/connect.spec.js
```diff
@@ -371,6 +382,37 @@ describe('React', () => {
       })
     })
 
+    it('should not refresh while tab is not visible and refreshHidden is unset', () => {
+      const interval = 100000 // set sufficently far out to not happen during test
+      const fethSpy = sandbox.spy(window, 'fetch')
```
ryanbrainard: spelling

test/components/connect.spec.js
```diff
@@ -371,6 +382,37 @@ describe('React', () => {
...
+        }
+      }
+
+      TestUtils.renderIntoDocument(
+        <Container />
+      )
+
+      expect(fethSpy.called).toEqual(false)
+
+      // trigger visivility change
```
ryanbrainard: spelling

test/components/connect.spec.js
```diff
@@ -371,6 +382,37 @@ describe('React', () => {
       })
     })
 
+    it('should not refresh while tab is not visible and refreshHidden is unset', () => {
```
ryanbrainard: can you add a test for `refreshHidden: true`?

package.json
```diff
@@ -51,6 +51,8 @@
     "react-addons-test-utils": "^0.14.0",
     "react-dom": "^0.14.0",
     "rimraf": "^2.3.4",
+    "sinon": "^1.17.2",
+    "visibilityjs": "^1.2.1",
```
ryanbrainard: shouldn't this be in `dependencies`?

ryanbrainard:
@hekike I tried this out locally, and it works and looks good overall, but left some comments for some minor changes. Thanks again for working on this.

test/components/connect.spec.js
hekike: done

src/components/connect.js
hekike: ok I'm waiting

src/components/connect.js
hekike: I understand and agree but seems to be that `this.state.mappings[prop]` doesn't have `fulfilled ` property. I need to dig deeper to understand the code and check for it. Could you help here?

hekike:
@ryanbrainard sorry for the late response. I made most of the requested changes. I also commented to te outstanding ones.

src/components/connect.js
ryanbrainard: Merged. Sorry for the delay.

XVincentX:
Is this still happening?

ryanbrainard:
Due to lack of recent activity on this PR, I'm closing it. If this is something still requested, please open a new PR.

PR #25, opened 2015-12-27
Update Readme

jwdotjs:
Replace instances of "updateState" with "updateStatus" in README.md

(Let me know if stripping the trailing whitespace is problematic, and I can update the PR)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 React Refetch
 =========================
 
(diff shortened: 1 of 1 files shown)


ryanbrainard:
Thanks! Not sure how all them at trailing white space got in there. I'm guessing it was from me, but I thought my editor would have cleaned that up for me on save. Perhaps it was from direct edits in GitHub.

PR #32, opened 2016-01-04
Apply defaults during coercion to consider for equality

### LINKED ISSUE #31: Request equality does not consider default request attributes
Requests are compared before defaults are applied, so if one request depends on an implicit default attribute and one explicitly declares the default attribute, they will be considered not equal. 

For example (reported in #26), the first `deviceMessageSearch` has an implicit `method : 'GET'`, but the second `deviceMessageSearch` has an explicit `method : 'GET'`. Even though the two are essentially equal, they are considered different requests and fetched again one time on `this.props.refreshMessages()`:

```jsx
export default connect((props) => {
  const deviceMessageUrl = `${process.env.API_URL}api/users/${props.params.userId}/device-messages`;
  return {
    deviceMessageSearch : deviceMessageUrl,
    refreshMessages : () => ({
      deviceMessageSearch : {
        url : deviceMessageUrl,
        method : 'GET',
        refreshing : true
      },
    }),
  };
})(DeviceMessages)
```

The fix is probably to break out the defaults from `buildRequest()` and apply them during request coercion so they are memoized and compared during the equality check.  

cc: @jwdotjs

ryanbrainard:
This moves the application of defaults out of `buildRequest()` and into a separate function that is called during request coercion. This allows the defaults to be considered for equality so that implicit and explicit request attributes are treated the same and fixes #31.

As part of fixing this, I realized that the `shallowEquals()` was broken for objects (became apparent once default headers were considered) and was missing tests. Somehow these got dropped in the fork from react-redux. This also fixes this issue.

cc: 
 - @jwdotjs: I confirmed this fixed your issue
 - @hekike: You may wish to rebase https://github.com/heroku/react-refetch/pull/23 on this and do your default handing in `applyDefaults()`
 - @jsullivan

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -49,6 +49,8 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
     invariant(isPlainObject(mapping), 'Request for `%s` must be either a string or a plain object. Instead received %s', prop, mapping)
     invariant(mapping.url, 'Request object for `%s` must have `url` attribute.', prop)
 
+    mapping = assignDefaults(mapping)
+
     mapping.equals = function (that) {
       if (this.comparison !== undefined) {
         return this.comparison === that.comparison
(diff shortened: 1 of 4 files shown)


hekike:
@ryanbrainard thanks to letting me know, I will!

ryanbrainard:
@jsullivan Can you review this when you get a chance?

ryanbrainard:

Issue #33, opened 2016-01-04
Identity Requests (value attribute)


ryanbrainard:
Add a `value` attribute to requests (aka mappings) the return the `value` in lieu of actually fetching data from `url`. Requests with `value` are not allowed to also have a `value`.

This will support composable transformations, like this (from #21):

```jsx
connect((props) => ({
  data: {
    url: "/api/data",
    then: (rawValue) => ({ 
      value: rawValue.real_data 
    })
  }
}))(Component)
```

and will also provide a way to create static `PromiseState`s, like this (from #19):

```jsx
connect((props) => ({
    project: `/api/projects/${props.id}`,
    updateProject: (data) => ({
        project: { 
          url: `/api/projects/${props.id}`, 
          method: "PATCH", 
          body: data,
          andThen: () => ({
            projectUpdated: { 
              value: true // some static value of your choice
            }
          })
        }
    })
}))
```

georgebonnr:
Trying refetch out tonight, I was just about to ask about this, to help with transformations.  Any timeline, or is this a request for a PR?

ryanbrainard:
@georgebonnr 

> I was just about to ask about this, to help with transformations.

Identity requests will allow for transformations of responses inside of `connect()`; however, you can use `PromiseState.then()` to do the same thing in the component. The only downsides of doing it that way are 1) gets applied for any prop change and 2) could clutter up the component if complex. Here's an example to transform the `usersFetch` to only include the verified users:

    this.props.usersFetch.then(users => users.filter(u => u.verified))

> Any timeline, or is this a request for a PR

I'd like to get to this soon, but we don't actually have a real need for it at this point with our own application of React Refetch, so that's slowing it down a bit. Also, there are a few different asks about transformations, handling different data types, etc, and I'd like to think about them holistically before rushing on any of them in piecemeal; however, I feel pretty confident about this one adding value with or without the others. I might even play around with it tonight since it should be fairly strait forward. With that said, anyone is welcome to open a PR for any of these enhancements.

ryanbrainard:
Implemented in #36

georgebonnr:
> Identity requests will allow for transformations of responses inside of connect(); however, you can use PromiseState.then() to do the same thing in the component.

Cool, yeah that's what I was planning on doing until identity is implemented – but no need now!  :clap:

ryanbrainard:
Merged #36 and released in 0.7.0-beta.0. Please kick the tires :)

[...]

Issue #18, opened 2015-12-21
Custom Request Post-Processing


bbirand:
I really like this library, and I'd like to try it for a data-intensive app. Currently, I use d3's `csv` method to do this:

```js
// (It's CSV, but GitHub Pages only gzip's JSON at the moment.)
d3.csv("large_dataset.csv", function(error, dat) {
   //work with dat
}
```

I implemented a version with react-refetch, that looks like this:
```js
let AnalysisRefetch = connect((props) => ({
    fetchedData: `large_dataset.csv`
}))(Analysis)
```

But when I look into `this.props.fetchedData`, the promise is rejected with the reason "SyntaxError: Unexpected token ,". 

I imagine this is because refetch was built for fetching JSON objects. But can I plug in a custom parser or so on the returned value? Or have access to the raw response, somehow? Using refetch will make things so much easier!

Thanks

bbirand:
I see now that this is the place where the magic happens:

https://github.com/heroku/react-refetch/blob/019a17057d7af6c22e4c9e160e77d08a11ec41c0/src/components/connect.js#L79

If you can point me in the right direction, I can try to give this a shot.

hnordt:
I would like to suggest something like this:

```js
connect(({ name }) => ({
  barFetch: {
    url: `/foos/${name}.csv`,
    dataType: 'csv' // maybe responseType?
  }
}))
```

ryanbrainard:
I'm all for accepting than just JSON, but I'd kind of like to have it based on the Content-Type header in the response rather than adding a new attribute. That might be wishful thinking since there's always going to be some API that doesn't have the right headers and anyway clients will need to send the Accept header as well. 

Another idea is to have customizable response processors. This is kind of possible now with PromiseState.then(), but since it's defaulting to JSON, it's too late once it gets to then(). 

So, given all this, I'm thinking we should use the response Content-Type if we can, fall through to JSON for backward compatibility, and allow it to be overridden by an attribute. Thoughts?

hnordt:
@ryanbrainard :+1: for Content-Type.

Please look at my enhancers idea: https://github.com/heroku/react-refetch/issues/9#issuecomment-166420907

hnordt:
@ryanbrainard my suggestion on https://github.com/heroku/react-refetch/issues/40#issuecomment-172942228 can handle this

ryanbrainard:
[...]

PR #36, opened 2016-01-11
Add Support for Identity Requests

ryanbrainard:
This implements "identity requests" as described in #33. See [Identity Requests: Static Data & Transforming Responses](https://github.com/heroku/react-refetch/blob/identity-requests/README.md#identity-requests-static-data--transforming-responses) for an explanation and example.

Note, this is not meant to handle the use case in #18 for post-processing (e.g. handling CSV responses) options. There is surely some overlap, but this is for transforming responses after any post-processing.

cc: @jsullivan @hburrows @georgebonnr @ashtuchkin @hnordt

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -159,51 +159,57 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
           window.clearTimeout(this.state.refreshTimeouts[prop])
         }
 
-        if (mapping.value) {
-          this.setAtomicState(prop, startedAt, mapping, PromiseState.resolve(mapping.value), null)
-          return
-        }
+        return this.createPromise(prop, mapping, startedAt).then(([ value, meta ]) => {
+          let refreshTimeout = null
+          if (mapping.refreshInterval > 0) {
+            refreshTimeout = window.setTimeout(() => {
+              this.refetchDatum(prop, Object.assign({}, mapping, { refreshing: true, force: true }))
+            }, mapping.refreshInterval)
+          }
 
-        const request = buildRequest(mapping)
-        const meta = { request: request }
+          if (Function.prototype.isPrototypeOf(mapping.then)) {
+            this.refetchDatum(prop, coerceMapping(null, mapping.then(value, meta)))
+            return
+          }
 
-        const initPS = mapping.refreshing ? PromiseState.refresh(this.state.data[prop], meta) : PromiseState.create(meta)
-        this.setAtomicState(prop, startedAt, mapping, initPS, null)
-
-        window.fetch(request).then(response => {
-          meta.response = response
-
-          return Promise.resolve(response).then(handleResponse).then(value => {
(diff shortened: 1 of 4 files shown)


src/components/connect.js
```diff
@@ -47,14 +47,15 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
...
+    invariant(mapping.url || mapping.value, 'Request object for `%s` must have `url` (or `value`) attribute.', prop)
+    invariant(!(mapping.url && mapping.value), 'Request object for `%s` must not have both `url` and `value` attributes.', prop)
 
     mapping.equals = function (that) {
       if (this.comparison !== undefined) {
         return this.comparison === that.comparison
       }
 
-      return [ 'url', 'method', 'headers', 'body' ].every((c) => {
+      return [ 'value', 'url', 'method', 'headers', 'body' ].every((c) => {
```
georgebonnr: Not 100% familiar with the code...  But if you have `value: doTransformationWork(rawValue)` in your mapping, that work will be done every time to produce the final value, so the equality check here doesn't prevent that work from being done?

That being said, seems like the primary use case for the value prop is in a `then` or `andThen` prop...  if it's being used in a 'top level' mapping, then why wouldn't you instead just apply the transformations to plain props directly?

Is that correct or am I missing something?  Going to be using this in .then so not really a problem.  But I guess I can also use my own memoization on that transformation function if I ever needed to.

ryanbrainard: @georgebonnr Correct, it will not prevent a function from being called to create `value`, but it will prevent the `PromiseState` from being updated with the same `value`. If you want to memoize the function result, it would need to happen outside of `connect()` or possibly use `comparison` to have finer grain control of when the `PromiseState` is updated, but not sure it would really help much for something just based on a top-level prop. As you mentioned, the typical case inside of `then` or `andThen` will stop the function from being called unnecessarily.

src/components/connect.js
```diff
@@ -158,6 +159,11 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
           window.clearTimeout(this.state.refreshTimeouts[prop])
         }
 
+        if (mapping.value) {
+          this.setAtomicState(prop, startedAt, mapping, PromiseState.resolve(mapping.value), null)
+          return
```
ryanbrainard: just realized this `return` here will block chaining off of an identity request. Not really sure what the use case would be for that, but for consistency, identity requests should still support chaining. need to fix.

ryanbrainard:
Just added a new [commit](d8b0e17) that breaks out `createPromise()` to treat identities as a resolved `Promise` and work the same as HTTP fetch for post processing. This abstracts things a bit and should make it easier to introduce generic/customizable request handling.

georgebonnr:
cool!  going to go ahead and point to this branch for now.

ryanbrainard:

PR #38, opened 2016-01-14
Add cause to errors to provide access to arbitrary fields

### LINKED ISSUE #37: Rejected promises don't get error messages in IE8
Hi,

I'm not sure if IE8 is supposed to be supported or not, but after polyfilling all of the things this library almost works perfectly.

The only issue I've encountered is that IE8 refuses to accept a message for the [error that is thrown when rejecting promises](https://github.com/heroku/react-refetch/blob/master/src/components/connect.js#L86-L88) because of the second `id` parameter. If you delete the `id` then everything seems to work fine.

As far as I can tell `Error` isn't supposed to have a second argument anyway so I'm not sure what the `id` is for, but it could be handled with a custom error type that does take a second argument - this could also be useful to support passing more than just a message on a server error.

ryanbrainard:
- Replaces the (erroneous) `id` field on errors with the full error in a `cause` field
 - Returns the error as a rejection rather than a throw in the Promise
 - Replaces the unhelpful stringifiation of error with a `message` or `error` field to just an empty string
 - Adds error handling tests

Fixes #37 

cc: @jsullivan @imjoehaines

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -3,6 +3,7 @@ import React, { Component } from 'react'
 import isPlainObject from '../utils/isPlainObject'
 import deepValue from '../utils/deepValue'
 import shallowEqual from '../utils/shallowEqual'
+import newError from '../utils/errors'
 import PromiseState from '../PromiseState'
 import hoistStatics from 'hoist-non-react-statics'
 import invariant from 'invariant'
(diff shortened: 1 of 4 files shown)


imjoehaines:
Just tested in Chrome & IE and it works fine. Thanks for fixing this so quickly!

ryanbrainard:

PR #41, opened 2016-01-19
Passthrough provided meta in identity requests

ryanbrainard:
Identity requests (i.e. requests with `value`) should also support `meta` to pass along to along in transformations; otherwise, this metadata gets lost and is not accessible by the component.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -26,7 +26,8 @@ Instead, it *returns* a new, connected component class, for you to use.
      - `andThen(value, meta): { prop: request, ... }` *(Function)*: returns an object of request mappings to fetch after fulfillment of this request but does not replace this request. Takes the `value` and `meta` of this request as arguments.
      - `andCatch(reason, meta): { prop: request, ... }` *(Function)*: returns an object of request mappings to fetch after rejection of this request but does not replace this request. Takes the `value` and `meta` of this request as arguments.
      - `value` *(Any)*: Data to passthrough directly to `PromiseState` as an alternative to providing a URL. This is an advanced option used for static data and data transformations. 
-
+     - `meta` *(Any)*: Metadata to passthrough directly to `PromiseState`. This attribute is only recognized if `value` is also provided.
+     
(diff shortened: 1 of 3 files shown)


src/components/connect.js
```diff
@@ -179,7 +179,7 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
         const onRejection = this.createPromiseStateOnRejection(prop, mapping, startedAt)
 
         if (mapping.value) {
-          const meta = {}
+          const meta = mapping.meta || {}
           this.setAtomicState(prop, startedAt, mapping, initPS(meta))
```
ryanbrainard: I went ahead and made this set the provided `meta` on the pending `PromiseState`; however, this is different behavior from the `value`, which is not set until the `PromiseState` is fulfilled. I'm not super happy about this, but I think it is the most expected behavior. We can think about setting `value` up front, but then it's different from regular URL-based requests. Feedback welcome.

cc: @jsullivan @hburrows

ryanbrainard:
Going to pull this in early for the beta, but please review async before this ships in 0.7.0

docs/api.md
```diff
@@ -26,7 +26,8 @@ Instead, it *returns* a new, connected component class, for you to use.
      - `andThen(value, meta): { prop: request, ... }` *(Function)*: returns an object of request mappings to fetch after fulfillment of this request but does not replace this request. Takes the `value` and `meta` of this request as arguments.
      - `andCatch(reason, meta): { prop: request, ... }` *(Function)*: returns an object of request mappings to fetch after rejection of this request but does not replace this request. Takes the `value` and `meta` of this request as arguments.
      - `value` *(Any)*: Data to passthrough directly to `PromiseState` as an alternative to providing a URL. This is an advanced option used for static data and data transformations. 
-
+     - `meta` *(Any)*: Metadata to passthrough directly to `PromiseState`. This attribute is only recognized if `value` is also provided.
```
ryanbrainard: See https://github.com/heroku/react-refetch/issues/42

PR #45, opened 2016-01-21
Allow to set the request mode in options

claudiuandrei:
Allow to configure request mode in case you want to make a request to a site that doesn’t support CORS.

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -71,6 +71,7 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
       {
         method: 'GET',
         credentials: 'same-origin',
+        mode: 'cors',
         redirect: 'follow'
       },
       mapping,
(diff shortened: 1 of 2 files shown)


claudiuandrei:
Closing, this may not be correct.

claudiuandrei:
Closing for now, there is an issue with parsing responses in the 'no-cors' mode. I need to investigate further.

Adam-Burke:
I'm looking for assistance in getting this library working with mode: 'no-cors'. Is this possible?

ryanbrainard:
@Adam-Burke Not currently possible, but should be very simple by just adding support for `mode` like this original PR did, but would need to be seriously rebased, so probably easier to start fresh. I'll open an enhancement request, and I'll try to get to it or would be happy to accept a new PR.

PR #48, opened 2016-02-01
Update whatwg-fetch version

thebuilder:
Allow whatwg-fetch to accept any version. Allows it to be used with the latest version 0.11.0.


ryanbrainard:
Might actually pull this out all together as a peer dep, but sounds good for now.

PR #64, opened 2016-02-23
Inherit comparison during transformation

### LINKED ISSUE #53: 'comparison' doesn't work when used along with then()
The comparison feature does not work if 'then' is also provided.

ryanbrainard:
This sets the `comparison` to value of parent so that `comparison`s will passthrough with `then` and `catch` transformations. The transformations can still provide their own `comparison`, but that is unlikely to ever be needed because the parent request is what triggers the fetch and needs to be compared.

Fixes #53 

cc: @eyalw @jsullivan

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -38,7 +38,7 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
     return mappings
   }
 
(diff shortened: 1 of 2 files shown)


jsullivan:
LGTM.

PR #65, opened 2016-02-23
Immediately resolve PromiseState for identity requests

### LINKED ISSUE #63: Static Data not immediately fulfilled?
Maybe I'm misunderstanding something, but I'd assume that [static data definitions](https://github.com/heroku/react-refetch#identity-requests-static-data--transforming-responses) like the following:

```es6
export default connect((props) => ({
  userFetch: { value: { id: 1, name: 'Tigger' }}
})(Component);
```

... would be available immediately, say, in the initial render:

```es6
render() {
  const user = this.props.userFetch.value;
}
```

However, it seems that on the initial render, `this.props.userFetch.pending` is `true`. I can't seem to get access to the static data until `componentWillReceiveProps`.

Shouldn't the `PromiseState` for a static data definition be immediately fulfilled?

ryanbrainard:
Not 100% (about 95%) sure we want to do this, but this fixes #63. It skips the `pending` state and resolves the `PromiseState` immediately for identity requests. See the [issue](https://github.com/heroku/react-refetch/issues/63) for some questions to consider. 

cc: @neezer @jsullivan

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -183,8 +183,7 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
         const onRejection = this.createPromiseStateOnRejection(prop, mapping, startedAt)
 
         if (mapping.value) {
-          this.setAtomicState(prop, startedAt, mapping, initPS(meta))
-          return Promise.resolve(mapping.value).then(onFulfillment(meta), onRejection(meta))
+          return onFulfillment(meta)(mapping.value)
         } else {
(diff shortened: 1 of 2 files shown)


neezer:
:+1:

...but I'm not the deciding vote. :wink:

ryanbrainard:
> ..but I'm not the deciding vote. :wink:

Giving people til EOD and then I'm merging this sucker :)

PR #68, opened 2016-02-24
Skip Response Body Parsing If Empty

### LINKED ISSUE #67: then/andThen with empty response body
With the following:

```es6
connect((props) => {
  fetchPrefs: '/api/prefs',
  deletePref: (id) => ({
    doDeletePref: {
      url: `/api/prefs/${id}`,
      method: 'DELETE',
      andThen: () => ({
        url: '/api/prefs',
        force: true,
        refreshing: true,
      })
    })
  }),
})(Component)
```

Invoking `deletePref()` never calls `andThen` on `doDeletePref`.

I've determined this is because my API returns a `204: No Content` with an empty response body. If I change the response to be a `200: OK` and return an empty JSON object, `andThen` fires normally and everything works.

I also tried returning a `200: OK` with an empty body, and the `andThen` did ***not*** fire.

So my guess is that **`andThen` only fires if the response body is not empty**. 

Personally, I think `andThen` should fire if the response was successful, regardless if the response body is present.

NOTE: I encountered the same behavior with `then`.

neezer:
Checks that the response 

- is not a `204: No Content`
- does not have a `Content-Length` header equal to `0`

If either of these are true, `handleResponse` skips parsing the response body.

Fixes #67

cc: @ryanbrainard

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -103,7 +103,12 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
   }
 
   function handleResponse(response) {
+    if (response.headers.get('content-length') === '0' || response.status === 204) {
+      return
+    }
+
     const json = response.json() // TODO: support other response types
+
     if (response.status >= 200 && response.status < 300) { // TODO: support custom acceptable statuses
       return json
     } else {
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -949,6 +949,70 @@ describe('React', () => {
       expect(decorated.refs.wrappedInstance.someInstanceMethod()).toBe(someData)
     })
 
+    it('should not parse the body if response is a 204', (done) => {
+      window.fetch = () => {
+        return new Promise((resolve) => {
+          resolve(new window.Response('', { status: 204 }))
+        })
+      }
+
+      @connect(() => ({ testFetch: `/empty` }))
+      class Container extends Component {
+        render() {
+          return <Passthrough {...this.props} />
+        }
+      }
+
+      const container = TestUtils.renderIntoDocument(
+        <Container />
+      )
+
+      const init = TestUtils.findRenderedComponentWithType(container, Container)
+      expect(init.state.data.testFetch).toIncludeKeyValues(
+        { fulfilled: false, pending: true, reason: null, refreshing: false, rejected: false, settled: false, value: null }
+      )
+
+      setImmediate(() => {
+        const fulfilled = TestUtils.findRenderedComponentWithType(container, Container)
+        expect(fulfilled.state.data.testFetch).toIncludeKeyValues(
+          { fulfilled: true, pending: false, reason: null, refreshing: false, rejected: false, settled: true, value: null }
+        )
+        done()
+      })
+    })
+
+    it('should not parse the body if response has Content-Length: 0', (done) => {
+      window.fetch = () => {
+        return new Promise((resolve) => {
+          resolve(new window.Response('', { status: 200, headers: { 'Content-Length': 0 } }))
+        })
+      }
+
+      @connect(() => ({ testFetch: `/empty` }))
+      class Container extends Component {
+        render() {
+          return <Passthrough {...this.props} />
+        }
+      }
+
+      const container = TestUtils.renderIntoDocument(
+        <Container />
+      )
+
+      const init = TestUtils.findRenderedComponentWithType(container, Container)
+      expect(init.state.data.testFetch).toIncludeKeyValues(
+        { fulfilled: false, pending: true, reason: null, refreshing: false, rejected: false, settled: false, value: null }
+      )
+
+      setImmediate(() => {
+        const fulfilled = TestUtils.findRenderedComponentWithType(container, Container)
+        expect(fulfilled.state.data.testFetch).toIncludeKeyValues(
+          { fulfilled: true, pending: false, reason: null, refreshing: false, rejected: false, settled: true, value: null }
+        )
+        done()
+      })
+    })
+
     // TODO
     //it('should not render the wrapped component when mapState does not produce change', () => {
     //  const store = createStore(stringBuilder)


src/components/connect.js
```diff
@@ -103,11 +103,15 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
   }
 
   function handleResponse(response) {
-    const json = response.json() // TODO: support other response types
-    if (response.status >= 200 && response.status < 300) { // TODO: support custom acceptable statuses
-      return json
-    } else {
-      return json.then(cause => Promise.reject(newError(cause)))
+    // TODO: support other response types
+    if (/^application\/json/.test(response.headers.get('content-type'))) {
```
neezer: I'm unsure how compatible this is with all derivative forms of the JSON MIME types (things like JSON API, etc.).

Also not sure if ancillary information like `;charset=UTF-8` is always guaranteed to come *after* the actual MIME type. Would appreciate some insight here.

test/components/connect.spec.js
```diff
@@ -949,6 +949,38 @@ describe('React', () => {
...
+
+      setImmediate(() => {
+        const fulfilled = TestUtils.findRenderedComponentWithType(container, Container)
+        expect(fulfilled.state.data.testFetch).toIncludeKeyValues(
+          { fulfilled: true, pending: false, reason: null, refreshing: false, rejected: false, settled: true, value: null }
+        )
+        done()
+      })
+    })
+
```
neezer: I did my best to strip this down to the essential test, but I feel like it could be stripped even more. Also the error message is not fantastic (buried in the diff provided by `toIncludeKeyValues`), so that could probably be improved.

ryanbrainard:
@neezer Thanks for the pull! 

I rather not base this on `Content-Type` because 1) it's common for 204 responses to still set `Content-Type: application/json` as a default header and 2) requiring `Content-Type: application/json` could be a breaking change. I was going to suggest that you key off of `Content-Length: 0`, but that does not work for `Transfer-Encoding: chunked`. I think there are two options here:

1. Special case the `204` status and do not attempt to parse the body. This should be fine according to the HTTP spec.
2. Always try to parse the body with an error handler for this case. This feels messy and error prone.

Since it's possible for other statuses to have empty bodies, it would probably be good to also check for `Content-Length: 0`. Together (`response.status === 204 || response.headers.get('content-length') === 0`), these should account for most cases and is non-breaking. I'd like to change how a lot of the media type handling works to support more then just JSON, and at that point, we could probably re-think how some of this works and be stricter about `Content-Type`.

test/components/connect.spec.js
```diff
@@ -12,9 +12,9 @@ describe('React', () => {
         fetchSpies.forEach((spy) => spy())
         return new Promise((resolve) => {
           if (request.url == '/error') {
-            resolve(new window.Response(JSON.stringify({ error: 'e', id: 'not_found' }), { status: 404 }))
+            resolve(new window.Response(JSON.stringify({ error: 'e', id: 'not_found' }), { status: 404, headers: { 'Content-Type': 'application/json' } }))
           } else {
```
ryanbrainard: you could add a `/empty` here instead of going it in the test. this fixture is all kind of messy (my fault), but at this way its consistent

test/components/connect.spec.js
ryanbrainard: Looks good. I know the tests are a bit verbose. Need to refactor some things. 

> Also the error message is not fantastic (buried in the diff provided by toIncludeKeyValues), so that could probably be improved.

What do you mean?

neezer:
> Together `(response.status === 204 || response.headers.get('content-length') === 0)`, these should account for most cases and is non-breaking.

@ryanbrainard Good points. I've updated this PR to reflect that, while also adding a new test case so that each condition is tested.

test/components/connect.spec.js
neezer: The failed assertion is

> Uncaught Error: expected { BIG OBJECT } to include { 'fulfilled': true }

... but the actual error is buried inside BIG OBJECT:

> ... reason: [SyntaxError: Unexpected end of input] ...

--

> Uncaught Error: expected { fulfilled: false, meta: { request: { _bodyInit: undefined, _bodyText: '', credentials: 'same-origin', headers: { map: { accept: [Object], 'content-type': [Object] } }, method: 'GET', mode: null, referrer: null, url: '/empty' }, response: { _bodyInit: '', _bodyText: '', bodyUsed: true, headers: { map: { 'content-length': [Object], 'content-type': [Object] } }, ok: true, status: 200, statusText: undefined, type: 'default', url: '' } }, pending: false, reason: [SyntaxError: Unexpected end of input], refreshing: false, rejected: true, settled: true, value: null } to include { 'fulfilled': true }

ryanbrainard: Oh, weird, I've never seen that. I guess the assertion diff facility is doing that

ryanbrainard:
:+1: Thanks!

test/components/connect.spec.js
ryanbrainard: FWIW I just tried running the tests locally with your fix commented, but I see the full object:

```
  1) React connect should not parse the body if response is a 204:
     Uncaught Error: expected { fulfilled: false, meta: { request: { _bodyInit: undefined, _bodyText: '', credentials: 'same-origin', headers: { map: { accept: [Object], 'content-type': [Object] } }, method: 'GET', mode: null, referrer: null, url: '/empty' }, response: { _bodyInit: '', _bodyText: '', bodyUsed: true, headers: { map: { 'content-type': [Object] } }, ok: true, status: 204, statusText: undefined, type: 'default', url: '' } }, pending: false, reason: [SyntaxError: Unexpected end of input], refreshing: false, rejected: true, settled: true, value: null } to include { 'fulfilled': true }
      at Function.assert (node_modules/expect/lib/assert.js:20:9)
      at setup.js:16:14
      at Array.forEach (native)
      at Expectation.toIncludeKeyValues (setup.js:11:27)
      at Immediate._onImmediate (connect.spec.js:977:48)
```

ryanbrainard:
Released: https://github.com/heroku/react-refetch/releases/tag/v0.7.0-beta.4

test/components/connect.spec.js
neezer: > ... but I see the full object

Yup, that's my point. My [previous post](https://github.com/heroku/react-refetch/pull/68#discussion_r53899088) was an abbreviation to help clarify my [initial comment](https://github.com/heroku/react-refetch/pull/68#discussion_r53890223), but I guess I missed the mark. :smile: 

All I was trying to say is that I think the failure should be clearer and less verbose, and take you right to the heart of the problem. It's a minor thing, though.

PR #71, opened 2016-02-24
key should be url

joerodrig3:
The README was just missing the key value for url


ryanbrainard:
This is ES6 object literal property value shorthand:
http://ariya.ofilabs.com/2013/02/es6-and-object-literal-property-value-shorthand.html

Issue #75, opened 2016-03-03
Supply custom fetch() implementation


passcod:
It would be great if a custom `fetch()` implementation could be supplied to React Refetch. My usecase would be not using a completely custom implementation (although I think that could be useful, too), but doing something like this:

```js
function customFetch (input, init = {}) {
  init.credentials = 'include'
  // do something on the input, too
  return window.fetch(input, init)
  .then(response => preprocess(response))
}
```

As well as implementing a kind of caching proxy that maintains a cache of responses and provides them directly from memory if Refetch asks for them based on some conditions (say, time since last request for that url, or `window.onLine === false`, or the heartbeat signal we use to check for _effective offlineness_, etc).

eyalw:
Have you tried this: https://github.com/werk85/fetch-intercept ?

passcod:
That's interesting, but not much more than my code block above. It acts globally, so it doesn't provide the ability to set interceptors or modifiers or proxies or whatnot _on a per-`connect` basis_, which is really what I wish for.

passcod:
It also doesn't play great with libraries or other parts of my code that use fetch but shouldn't be affected by the modifications I make for react-refetched components.

ryanbrainard:
Originally, I was pretty opposed to having something like this, but I'm starting to warm up the idea especially as since there's conflicting ideas around how pre and post processing should work and having more flexibility for isomorphic fetch and other fetch implementions. How are you imagining the you'd configure something like this? Some kind of fetch-provider you'd pass in as an argument to connect()?

passcod:
<sup>(Yes, I was reluctant at first, seeing replacing fetch as a nasty hack, but with further thought, the `fetch` interface is quite neat and simple, and I can definitely see a small ecosystem of stack-agnostic libraries that override `fetch`. Like gulp and node streams ubiquitised the Pipe Transform as building block for anything, and React got the `connect()` pattern, and ES7 brings `@decorators`, and express/rack made server middleware essential, so could `fetch` middleware become another such pattern and building block. Or maybe that's just wishful thinking.)</sup>

Ok, that _is_ tricky. I originally knee-jerked like "just add it to the options obj— ah" but there's no options object. Your API is really clean! :+1: 

Just for arguments sake, we _could_ add it to the main argument but uuugh so ugly:

```js
connect(props => ({
  fetchProvider: customFetch,
// or maybe:
  _fetch: customFetch,
// yeah, nah.
  usersFetch: {
    url: `/users`,
[...]

Issue #60, opened 2016-02-20
Enable standard use of the ref prop.


eyalw:
Instead of using a string value for the ref property, using the more modern method using the function value will enable regular use of the ref property from the parent of the wrapped component.
**today**:
https://github.com/heroku/react-refetch/blob/master/src/components/connect.js#L138 )

**suggested:**
![image](https://cloud.githubusercontent.com/assets/608669/13199018/63205690-d821-11e5-84fe-6b781409c606.png)

ryanbrainard:
This was lifted directly from react-redux during the fork. I would not be opposed to updating this, but would like to do so in a backward compatible way. It would also be nice if it could continue to work the same way as react-redux (perhaps two PRs are needed?)

eyalw:
so I guess we check the type of the ref prop (string/func) and act accordingly

ryanbrainard:
Related: is there any good reason react-redux requires `withRef: true` be set explicitly? Would it hurt anything to just make the ref always available?

eyalw:
I can't think of a reason why they require this...

nfcampos:
there is one reason why you don't want withRef to always be true (or even default to true), stateless functional components don't have refs, so if you wrap one of those you can't provide a ref

Issue #40, opened 2016-01-19
Offline Support


hnordt:
I'm working on a project that needs offline support and I don't want to use Flux, because all state is saved on server-side and I don't wanna duplicate/manage local state.

Would be awesome if we can provide our own implementation of `handleResponse`, so we can implement something like this:

```js
// myOwnHandleResponse.js

import localforage from 'localforage';

const getRequestId = request => {
  // TODO: md5 or something like that :P
  return `${request.url}.${request.method}.${JSON.stringify(request.headers)}`;
};

// On response success update offline cache and return json
// On response failure warn and return the cached json
const handleResponse = (request, response) => {
  const requestId = getRequestId(request);
  if (response.status >= 200 && response.status < 300) { // Or response.ok
    const json = response.json();
    // localforage.setItem() resolves the value that was set :)
    return localforage.setItem(requestId, json).catch(error => {
      console.warn('Failed to update offline cache', { requestId, request, error });
      return json;
    });
  } else {
    return localforage.getItem(requestId);
  }
};

export default handleResponse;
```

We'll need to implement a timeout in React Refetch and implement a way to handle POST/PUT/DELETE requests.
It's just the initial idea. Toughts?

ryanbrainard:
> a way to handle POST/PUT/DELETE requests.

Set the `method` in the request.

ryanbrainard:
I'm still idly noodling on how to handle custom request and response handling. There's several way to do it, but so far not totally satisfied with something that both retains reasonable defaults and gives users enough power. Good to see more use cases like this though.

hnordt:
@ryanbrainard it's not so simple, offline POST/PUT/DELETE requests need to be queued, so we can dispatch them when we back online. We'll need to do something like:

```js
const handleRequest = request => {
  if (request.method === 'POST' && isOffline) {
    return localforage.setItem('queue', [...queue, request]);
  }
  // ...
};
```

hnordt:
@ryanbrainard,

Something that is simple and very flexible:

Change:

```js
return fetched.then(handleResponse).then(onFulfillment(meta), onRejection(meta))
```

```js
function buildRequest(mapping) {
  return new window.Request(mapping.url, {
    method: mapping.method,
    headers: mapping.headers,
    credentials: mapping.credentials,
    redirect: mapping.redirect,
    body: mapping.body
  })
}
```

```js
function handleResponse(response) {
  const json = response.json() // TODO: support other response types
[...]

Issue #11, opened 2015-12-18
Switch to isomorphic-fetch


nicholaswyoung:
The module, as it sits, is almost perfect for my use case. However, I need to render components server side too. Would you be open to a PR that replaces `whatwg-fetch` with a similar, standards compliant polyfill that also works in Node?

ryanbrainard:
I'm not opposed to it, but just wondering if it would be better to allow applications to provide their own `fetch` polyfill. Perhaps it should just be a dev dependency in this library and then it would be up to applications to choose what works best for them? Also, `whatwg-fetch` will not load if you already have another `fetch` loaded already, so you could probably still use it as-is today and just make sure your `fetch` is loaded first. I do not claim to be an expert at all on JavaScript modules, imports, and polyfills, so  I'm all ears if someone has a better idea how best to do this.

nicholaswyoung:
@ryanbrainard That's an even better idea. The big issue of mine was that Node doesn't recognize `window`, so we'll need to decide how that dependency is passed in. That's what originally led me down this path.

I'll dig into this and see if I can come up with a workable solution.

wilmoore:
@nicholaswyoung 

> ...The big issue of mine was that Node doesn't recognize window, so we'll need to decide how that dependency is passed in. That's what originally led me down this path.

You might check out: https://www.npmjs.com/package/global

hnordt:
@ryanbrainard I'm just wondering why `whatwg-fetch` is a peerDependency and not a dependency?

ryanbrainard:
@hnordt The idea was to allow applications to have more freedom to choose which version of `whatwg-fetch` so it would not conflict if it was already a dependency of the application. Although, to be honest, I'm kind of an npm noob, so if someone knows a better way to do this, please let me know. If we can get away with just making it a devDependency for tests and have applications use whatever polyfill they want (or nothing at all if they don't care about cross-browser support) in production would be ideal. How do other npm modules usually deal with polyfill dependencies like this?

wilmoore:
> How do other npm modules usually deal with polyfill dependencies like this?

Other npm packages omit the dependency from `dependencies`, but put it in `devDependencies` for tests. The user is then responsible for installing the actual dependency into their own project.

ryanbrainard:
> Other npm packages omit the dependency from dependencies, but put it in devDependencies for tests. The user is then responsible for installing the actual dependency into their own project.

[...]

PR #76, opened 2016-03-05
Add option to specify fetch implementation

passcod:
PR for #75.

I think that covers it, it's a pretty simple feature :) The core focus of this PR works, and perfectly: one can now override the fetch implementation on a `connect` call basis. But details remain:

I'm not sure about the chaining configuration API. To be exact: I like the API, but I'm not too happy about how I implemented it, but that's what feedback is for, right? The way it works is it preserves the existing options hash so the implementation is simple, but provides the chaining API on top. The way I did the chaining means you can only use a single `.with*` to set the relevant option:

```js
// Possible due to the implementation, but discouraged:
@connect({
  usersFetch: '/example'
}, { withFetch: fetch })
class Foo extends Component {}

// Chaining (recommended for public use):
@connect.withFetch(fetch)({ ... })
class Foo extends Component {}

// Multiple chaining (doesn't work with current PR):
@connect.withFetch(fetch).withRef()({ ... })
class Foo extends Component {}

// "Detached" style (functionally the same thing)
const refetch = connect.withFetch(fetch).withRef()

@refetch({ ... })
class Foo extends Component {}
```

I have (as you can see above) added the `withFetch` option chainer. If you think that is best done in future work detached from this PR, I will remove it.

I have added a test for this. I have _not_, however, changed the rest of the tests around to use `withFetch` to do dependency injection instead of overwriting `window.fetch`, which is something @ryanbrainard [mentioned](https://github.com/heroku/react-refetch/issues/75#issuecomment-192032812) as a potential outcome of this PR.

I have modified the `wrappedInstance` exception message to mention using `.withFetch()` instead of `{ withFetch: true }`. I should probably revert that? (As it may be a breaking change, and not particularly necessary to do in this PR.)

Really, the only thing bugging me about this is the chaining option impl. If someone has an idea on how to better do it…


nfcampos:
well, the easiest way to allow chaining is to not need it
```js
connect.with({fetch: fetch})
connect.with({withRef: true, fetch: fetch})
```
but i can't say i'm too happy about this either

nfcampos:
and one way i can think of to get rid of the options hash is to keep track of it in a closure
```js
function connectFactory(options) {
  function connect(mapPropsToRequestsToProps) {
    ...
  }
  connect.withFetch = function(fetch) {
    return connectFactory(Object.assign({}, options, {fetch: fetch}))
  }
  connect.withRef = function(withRef) {
    return connectFactory(Object.assign({}, options, {withRef: withRef}))
  }
  return connect
}
const defaults = {
  fetch: window.fetch,
  withRef: true
}
export default connectFactory(defaults)
```
this actually also allows for chaining `connect.withFetch(fetch).withRef(false)` unless i did some horrible mistake
(you'd want to use _.assign instead of Object.assign to avoid messing with polyfills)

passcod:
Looks to work fine, except I now have half of the tests failing and it's the _wrong_ half (the withFetch and chaining tests pass fine). Investigating.

nfcampos:
let me know if you want another pair of eyes looking into the test failures :)

passcod:
Sure! Here's [a gist](https://gist.github.com/passcod/23a8ec319b62ebbddd30) with the test fails. I've pushed the failing code on this PR (I'll squash when it's all fixed :) It's past midnight for me, going to adjourn until tomorrow.

nfcampos:
I'll take a look

src/components/connect.js
```diff
@@ -17,9 +17,32 @@ function getDisplayName(WrappedComponent) {
...
+  fetch: window.fetch,
+  withRef: false
+}
+
+export default connectFactory(defaults)
+
+function connect(mapPropsToRequestsToProps, options = {}) {
   const finalMapPropsToRequestsToProps = mapPropsToRequestsToProps || defaultMapPropsToRequestsToProps
-  const { withRef = false } = options
+  const { fetch, withRef } = options
```
nfcampos: @passcod if you move the defaults inside connect() all the tests pass
this line becomes:
```js
const { fetch=window.fetch, withRef=false } = options
```
remove the defaults outside of connect and do
```js
export default connectFactory()
```
it's something to do with window.fetch not being defined when the module is imported, but being available at the time the connect function is called, that's why inside connect() works but outside it doesn't

src/components/connect.js
```diff
@@ -17,9 +17,32 @@ function getDisplayName(WrappedComponent) {
...
 let nextVersion = 0
 
-export default function connect(mapPropsToRequestsToProps, options = {}) {
+function connectFactory(options) {
+  function connectImpl(map) {
+    return connect(map, options)
+  }
+
+  connectImpl.withFetch = function (fetch) {
+    return connectFactory(assign({}, options, { fetch: fetch }))
```
nfcampos: we might want to add some check here to make sure the passed argument at least is a function

passcod:
All fixed up, and squashed into one commit. @ryanbrainard review?

src/components/connect.js
```diff
@@ -17,9 +17,36 @@ function getDisplayName(WrappedComponent) {
...
+  return connectImpl
+}
+
+export default connectFactory()
+
+function connect(mapPropsToRequestsToProps, options) {
   const finalMapPropsToRequestsToProps = mapPropsToRequestsToProps || defaultMapPropsToRequestsToProps
-  const { withRef = false } = options
+  const {
+    fetch = window.fetch,
```
ryanbrainard: wonder if we should do just `fetch` here so that if someone is using [`isomorphic-fetch`](https://github.com/matthew-andrews/isomorphic-fetch), (I think) it will work out of the box.

src/components/connect.js
```diff
@@ -17,9 +17,36 @@ function getDisplayName(WrappedComponent) {
...
+    withRef = false
```
ryanbrainard: i think we can default this to `true` or probably even just remove it as an option, unless someone has a good reason you'd want to be able to not have a ref. See https://github.com/heroku/react-refetch/issues/60#issuecomment-187852641

src/components/connect.js
ryanbrainard: also, should we call this `fetch` const something different to avoid confusion w/ the global `fetch`?

src/components/connect.js
nfcampos: see my comment in that issue for a reason not to have a ref

src/components/connect.js
nfcampos: yes i agree with dropping the `window` part (one difference is it will throw a ReferenceError if it isn't defined, ie. if the browser doesn't have it and the user forgot to include a polyfill, but this is probably a plus actually)

nfcampos: about the name I'm not sure, you're right it might be confusing for someone dropping in on the code without context, but would a comment suffice?

ryanbrainard: Perhaps add an `invariant` check before this so the developer gets a better error message in console?

nfcampos: yeah an `invariant` is indeed better than the ReferenceError

src/utils/assign.js
```diff
@@ -0,0 +1,13 @@
+export default function assign(obj, ...sources) {
```
ryanbrainard: what is this achieving over the standard `Object.assign`?

nfcampos: we can only be sure Object.assign is available if a polyfill is included, no?

ryanbrainard: Yeah, I guess browser support isn't as universal as I thought: 
https://kangax.github.io/compat-table/es6/#test-Object_static_methods_Object.assign

nfcampos: but actually i think it'd be better to use the assign function from lodash or some other library rather than maintaining our own assign method

src/components/connect.js
```diff
@@ -1,5 +1,5 @@
-import 'whatwg-fetch'
```
ryanbrainard: can we move `'whatwg-fetch'` from `peerDependencies` into `devDependencies` and make a note in the readme (and an invariant in code) to make it the app's responsibility to provide fetch? This should also deliver #11 since app's can just use `isomorphic-fetch` instead.

src/utils/assign.js
ryanbrainard: I'd rather not pull in another dep

src/components/connect.js
ryanbrainard: Related: there are still uses of `window.Request`. Can this just be 'Request`? Any reason for this to be customizable as well?

nfcampos: yep just `Request` should be enough. customising `Request` sounds like a bad idea since it's the Request object is used internally by fetch, no?

ryanbrainard:
@passcod Looking good so far! I just tested it out on our internal project that uses RR, and it worked without modification and I tried a simple wrapper for a connect and it worked. Left a few minor comments, but otherwise, this is looking good. Can you also add an example into the readme about this? Perhaps also something about what apps should or should not do in their custom fetch?

src/utils/assign.js
passcod: Yeah @nfcampos mentioned the polyfill thing, but lodash wasn't already included, so I wrote that. I just checked the [`Object.assign` MDN page](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign), and it shows an ES5 implementation which I believe is functionally equivalent to mine. We can revert to native `Object.assign`, as the code was already peppered with calls to it before this PR, though.

src/components/connect.js
passcod: @ryanbrainard I left as-is to avoid _too_ much scope creep on this PR.

nfcampos:
speaking of uses, i just realised my main usecase for a custom fetch won't work with this:
(supporting long polling)
```js
const customFetch = (...args) => {
  return fetch(...args)
    .then(resp => resp.json())
    .then(data => {
      if (data.task)
        return pollUntilTaskCompletes(data)
      else
        return data
    })
}
// pollUntilTaskCompletes polls for completion, returning a promise that resolves once the task is complete
```
This looks like the kind of thing one might do with a custom fetch.
However, since I need to consume the json of the request when `react-refetch`'s handleResponse() tries do the same it will fail...
If a custom fetch can't be used for request post-processing then exactly what can one do with this?
Actually, @passcod 's example on issue #75 won't work either (assuming it also needs to consume the json to process the response...

passcod:
@nfcampos You'll need to return a Response from within your last `.then()`. Given Responses can be constructed from user code, this is fine. Modify the data, then construct a new Response from the original and the new data. Then that will work fine with React-Refetch as you'll be respecting the `fetch` API as spec'd.

nfcampos:
you can only call `.json()` on a `Response` instance once, so I'd have to do something like
```js
  fetch(...args)
    .then(resp => [resp, resp.clone().json()])
    .then(([originalResponse, data]) => {
      if (data.task)
        return pollUntilTaskCompletes(data)
      else
        return originalResponse
    })
  // and remember to the same dance inside pollUntilTaskCompletes()
```
which I think we can all agree is pretty terrible...

src/components/connect.js
ryanbrainard: > about the name I'm not sure, you're right it might be confusing for someone dropping in on the code without context, but would a comment suffice?

Yeah, you're probably right. Leave it as is.

> yep just Request should be enough. customising Request sounds like a bad idea since it's the Request object is used internally by fetch, no? I mean, Request is basically an implementation detail of fetch(), so I don't really see anything that could be achieved with a custom Request that you couldn't do by supplying a custom fetch

+1

src/components/connect.js
ryanbrainard: Sounds good. We can do that in a second pass.

src/utils/assign.js
ryanbrainard: > We can revert to native Object.assign, as the code was already peppered with calls to it before this PR, though.

Yeah, let's revert to using `Object.assign`. This lib already expects it. We should just document all the polyfills expected: Object.assign, ES6 promises, fetch

ryanbrainard:
Now that we have a nice way to chain options onto `connect`, we could add nicer ways to set defaults and post-processing to avoid conflicts with the built-in JSON processing. As I've mentioned before, I'd really like to have declarative ways to set those kinds of things, but perhaps this can be the lower level option and then there's a higher level declarative thing later on. Either way, I think the way it is now still adds value, and more can be added later (i.e. don't add it to this PR :smile:)

passcod:
For the possibility of modifying Responses, it's not *that* terrible. One could write wrappers that take care of those details (that's my intention!).

nfcampos:
by terrible I mean it's error prone and not intuitive (how many people know you can only consume the body of a `Response` once?). So what is a good, non-hacky, use case for this?
- `Response` processing would certainly be better handled by allowing a custom handleResponse()
- `Request` pre-processing maybe?

src/components/connect.js
passcod: Ah, hmm, I'd missed the call to `window.Request` directly. If a completely custom `fetch` is provided, it would be reasonable to expect a similarly completely custom Request. Annoyingly, there's no way to derive `Request` from `fetch`. We could have a separate `.withRequest(Request)` to not use the global Request, or add it as optional parameter to `withFetch`.

The `fetch` function signature can be any of:

```js
fetch(req: Request)
fetch(url: String)
fetch(url: String, options: PlainObject) // same args passed to Request constructor
fetch(req: Request, options: PlainObject) // options here override Request ones 
```

So one can easily imagine switching between those forms to be done upon `firstArg instanceof Request`, where `Request` is _internal to the fetch implementation_. That's how [github/fetch aka whatwg-fetch](https://github.com/github/fetch/blob/master/fetch.js#L330) does it, for example. Also [this other fetch polyfill](https://github.com/camsong/fetch-ie8/blob/master/fetch.js#L330). Heck, I just tried replacing `window.Request` in my browser (which has a native `fetch`), and from the behaviour, it seems it does the same thing.

Thus, using `window.Request` with a non-`window` fetch is probably a no-go.

passcod:
Request preprocessing, definitely. `fetch`-level custom caching (construct and return a Response from memory or, say, IndexedDB instead of trying the network. A kind of page-local Service Worker that works today, if you will). Response post-processing with caution (or with something that takes care of it). As @ryanbrainard asked, I'll add something to the Readme about `withFetch()`... mentioning these caveats there is probably a good idea, right?

ryanbrainard:
> mentioning these caveats there is probably a good idea, right?

Yes, and probably something like "This is an advanced feature. Use existing declarative functionality wherever possible, but this is provided as a way to control the `fetch` implementation when necessary. Please be aware that changing the `fetch` implementation could conflict with built-in current or future functionality."

nfcampos:
I feel that this API would be more useful if it embraced the fact that if you're replacing the transport (fetch) it's extremely likely you'll want to replace some other part of the network logic, namely request pre- and post-processing.
how about this?
```js
connect.withTransport((defaultBuildRequest, defaultHandleResponse) => {
  return function(mapping) {
    // do whatever you want
    return data
  }
}
```
in which the default implementation is
```js
connect.withTransport((defaultBuildRequest, defaultHandleResponse) => {
  return function(mapping) {
    return fetch(defaultBuildRequest(mapping).then(defaultHandleResponse)
  }
}
```
Otherwise, since there's practically no use for replacing `fetch()` in which you don't need to replace either `buildRequest` or `handleResponse` you'll always end up working around the default implementations of these two.

src/components/connect.js
ryanbrainard: I like that we are dropping the options hash here, but I'm slightly worried that that is a breaking change. What do you think about still allowing it in the options hash with something that emits a warning? We can leave it in there for one release and then pull it out.

ryanbrainard:
@nfcampos I understand what you're saying, but wha about cases where you only want to replace the pre or post processing? That means you'd be forced to re-write half the default implementation (even though it is only one line). What would you think of having these as three separate options (i.e. `withBuildRequest`, `withFetch`, `withHandleResponse`)?

nfcampos:
in which:
- buildRequest is `mapping -> Request`
- fetch is `Request -> Response`
- handleResponse is `Response -> value`

something like this?

ryanbrainard:
Yes, except `handleResponse is Response -> Promise` since the standard `onFulfillment` and `onRejection` handlers would still need to be chained after whatever is overridden.

nfcampos:
exactly

passcod:
Sounds good. Do we want to do this in this PR or in another one?

ryanbrainard:
> Do we want to do this in this PR or in another one as follow up?

Your choice. I don't want to cut a release until they are all in, but your choice on how to break up the PRs.

ryanbrainard:
Also, what do you think of the names `withBuildRequest` and `withHandleResponse`? They feel slightly clunky to me. Anyone have more concise names to express what these do?

nfcampos:
(note for the docs: all requests go through this pipeline of three functions except for identity requests, which bypass all three, returning value immediately)

nfcampos:
`withPre`, `withPost`? too cryptic?

src/components/connect.js
passcod: Ah yes. We could make it transparently merge options and display a warning:

```js
  function connectImpl(map, callOptions = {}) {
    console.warn(...)
    return connect(map, Object.assign(options, callOptions))
  }
```

src/utils/assign.js
passcod: Ok :)

ryanbrainard:
We should also consider how `assignDefaults` plays in with all this. That happens during mapping coercion, but would probably would want to be updated with (or instead of) the request builder. I don't really want to expand the scope too much here, but something we should consider before releasing this.

ryanbrainard:
> `withPre`, `withPost`? too cryptic?

Yes, pre-what? post-what?

nfcampos:
`preFetch`,`fetch`, `postFetch`?
`assignDefaults` should just be part of the default buildRequest no?

passcod:
That new proposal of three functions also takes care of the `window.Request` call [as discussed above](https://github.com/heroku/react-refetch/pull/76#discussion_r55280178), very good.

nfcampos:
`withBeforeFetch`, `withFetch`, `withAfterFetch` ?

ryanbrainard:
> assignDefaults should just be part of the default buildRequest no?

It's probably doing too much now. The request-specific parts (`headers`, `method`, `credentials`, `redirect`) of it probably should be moved into `buildRequest`, and `comparison` and `meta` should be in `coerceMapping`.

nfcampos:
@ryanbrainard yep, i agree, that division makes sense to me

passcod:
`withBeforeFetch` doesn't really read well. Maybe either `.beforeFetch()`, leaving off the `with` prefix, but that runs into the problem of not really describing what's _happening_ before fetch. So instead perhaps `withRequestBuilder()`, as a rewording of the original?

nfcampos:
and `withReponseHandler()`?

ryanbrainard:
Not, sure I really like them as nouns. Sounds like they take objects (and it's `withFetch`, not `withFetcher`). Probably just go back to the original `withBuildRequest`, `withFetch`, `withHandleResponse`. Slightly clunky, but they are the clearest. Good we talked through options though

passcod:
Good point on the `withFetcher`. Anyway, these don't look like words to me anymore, ugh. I'm going to take a break and come back to this later today. As a side note, I won't be able to work on this from Thursday until Monday or Tuesday due to surgery so if I don't respond in that time that's why ;)

ryanbrainard:
> I'm going to take a break and come back to this later today. As a side note, I won't be able to work on this from Thursday until Monday or Tuesday due to surgery so if I don't respond during that period that's why ;)

Sounds good. Good luck on the surgery!

nfcampos:
@passcod Good luck!

passcod:
Remains to do:

- [ ] Readme blurb about all these new shiny options, including caveats
- [ ] Mention that `Object.assign`, `Promise`, `fetch` have to be polyfilled by the user, if necessary.
- [ ] Test "Missing available fetch impl" invariant
- [ ] Test deprecated options usage warning
- [ ] Test isomorphic fetch: stub `global.fetch` and `self.fetch` instead of `window.fetch` to verify it picks it up
- [ ] Test isomorphic by e.g. taking out `window`
- [ ] Implement `withBuildRequest()`
- [ ] And test
- [ ] Implement `withHandleResponse()`
- [ ] And test
- [ ] Do [something](https://github.com/heroku/react-refetch/pull/76#issuecomment-193496802) about `assignDefaults`
- [ ] Clean up, squash, final checks, ask for review

ryanbrainard:
I've been thinking about the `assignDefaults` problem. In actual use, when
I have wanted to have common default for all mappings, I would just want to
be working with the mapping objects, not building out the full request.
Also, I was setting things like `comparison`, which would be outside the
request itself. Not exactly sure what the answer is, but the way we are
planning to do `buildRequest` would both be making me do too much and not
giving me enough control. Thoughts?

nfcampos:
Two thoughts:
- for `defaults` would it be enough to just provide a `withDefaults()` method that takes an object of defaults that is later used somewhat like so: `Object.assign({}, defaults, mapping)` (with some special handling to also merge the headers object, etc. For the use cases I can think of this sounds like enough, do you see any advantage to actually overriding the `assignDefaults` function?
- are four customisation hooks too much? Or can we just assume they are for specialised use and most people will use none or just `connect.withDefaults()` so it's fine?

ryanbrainard:
> for `defaults` would it be enough to just provide a `withDefaults()` method ...

Yes, I think that's how it should work, but the question is should it happen inside of `buildRequest` or before it? I'm thinking before it.

> are four customisation hooks too much?

Perhaps. I was thinking about this, and realizing that many times they will be used in together. For example, if you want to fetch a CSV file, you'll probably override defaults and have a custom response handler, or if you want to pull from offline storage, you'd override fetch and probably the other handlers. Because of this, perhaps, we should still think of them as separate, but part of something more cohesive (like a "profile" or something of related options). Perhaps just going back to treating them as an object like `connect.with({withRef: true, fetch: fetch})` would achieve this. The other reason I like it as an object is that these could be attributes pushed down into the mapping (and we just happen to be allowing them to be set at the `connect` level here). See https://github.com/heroku/react-refetch/issues/40#issuecomment-172949423 for some of my previous thoughts on this and why I was hesitant to go this route. I don't want to change to course here too much on this PR, but we should think about how these can also be set at the individual mapping level, and that might inform us to treat them as  attributes in an object.

passcod:
Could always do both: `withRef`, `withFetch`... and a convenience `.with({ ... })` that takes an object and maps it to the chained methods in one go.

nfcampos:
I kind of like the symmetry of the API as you proposed with `connect.with({})`. It sounds like a good idea to be able to do something like:
```js
@connect.with({comparison: ..., credentials: ..., fetch: () => {...}})(props => ({
  usersFetch: 'some/url',
  otherFetch: 'some/other/url',
})
class Blah extends Component {
  ...
}
```
and know that any keys provided to `with()`will simply act as defaults for the mappings

ryanbrainard:
@passcod We could, but that would more methods for every option. I could be getting ahead of myself here, but I'm thinking is something like this where `buildRequest`, `fetch`, and `handleResponse` are just options in the mappings like everything else AND any option in the mapping can be set in a default. Something like this (notice the connect-level and mapping-level `handleResponse`s):

```jsx
connect.defaults({
  buildRequest: ...,
  fetch: ...,
  handleResponse: ...,
  refreshing: true,
  headers: {
    'Custom-Header': 'true'
  }
}).(({ params: { userId } }) => ({
    userFetch:    `/users/${userId}`,
    likesStatusesFetch: {
      url: `/users/${userId}/likes`,
      handleResponse: ...
    }
}))(Profile)
```

This example looks a bit ugly, but if these things were extracted out, I think the overall API work and is quite nice. Again, I could be getting ahead of myself here and blowing out the scope of this PR (sorry!), but I at least wanted to get your thoughts on this, and if we like it, can iterate toward it.

ryanbrainard:
@nfcampos ha, I think we're saying the same thing in https://github.com/heroku/react-refetch/pull/76#issuecomment-194539138 and https://github.com/heroku/react-refetch/pull/76#issuecomment-194542891 :) Good timing on our comments too!

ryanbrainard:
oops, had a an extra `.` in my example. fixed it up

nfcampos:
@ryanbrainard yep exactly the same thing :)

ryanbrainard:
@nfcampos In mine, I changed the method name from `with` to `defaults`. Seemed clearer to me. Like that change?

Also, are you on board with the proposal overall? It sounds like you are...

nfcampos:
yep `defaults` is more obvious, so better :)
Yes, I'm on board with this, it creates all (all I can think of at least) the customisation hooks that seem useful and does it in a clear way that works both at the connect-level as well as at the individual-mapping level

passcod:
So [surgery went well](https://medium.com/@passcod/i-ve-seen-and-read-a-few-pieces-and-blogs-about-how-icl-goes-the-before-the-after-the-long-view-16159a19dbc6) and although I'm not going to actually work on this until the weekend, I wanted some clarification:

1. Is _every_ property that can be set in `defaults()` also available in the individual fetch blocks, or are there things that are `defaults`-only (or fetch-block-only)?
2. The list of properties that can be in `defaults`:
    - `buildRequest`
    - `fetch`
    - `handleResponse`
    - `refreshing`
    - `headers`
    - `comparison`
    - `credentials`
    - `method` ?
    - `redirect`
    - `refreshInterval`
    - `force`
    - `then`, `catch`, `andThen`, `andCatch` ? (should those be replaced or added onto if specified in both the `defaults` and the fetch block?)

Any I've missed?

As an addition, although this should probably be another PR done later, how about a `urlBase` or `urlPrefix` default so we can avoid repeating the same thing over and over?

```js
@connect.defaults({ urlBase: '/api/v3' })(props => ({
  usersFetch: '/users',
  projectFetch: `/projects/${props.projectId}`
}))

// instead of

@connect(props => ({
  usersFetch: '/api/v3/users',
  projectFetch: `/api/v3/projects/${props.projectId}`
}))
```

ryanbrainard:
Glad to hear it went well!

Yes, I'm thinking everything should be able to be set at both the connect and mapping levels. Some of them probably don't make much sense as defaults, but I'd rather not special case anything if possible. 

I like the idea of a base URL too! That is especially helpful with isomorphic fetch because it requires absolute URLs on the server side. Yes, probably a separate PR.

ryanbrainard:
@passcod Just wanted to check in on this PR. Looks like you added some commits, so guessing you're feeling better :) How close do you think this is from being ready for review and merge?

passcod:
Probably this weekend :)

ryanbrainard:
@passcod SGTM

passcod:
Checklist:

- [ ] Implement and test (see checklist on https://github.com/heroku/react-refetch/pull/76#issuecomment-195805510)
- [ ] Readme blurb, including caveats and usage guidelines for `fetch`
- [ ] Mention that `Object.assign` (also see #82), `Promise`, `fetch` have to be polyfilled by the user, if necessary.
- [ ] API docs for the new `defaults`
- [ ] Test "Missing available fetch impl" invariant
- [ ] Test deprecated options usage warning
- [ ] Test isomorphic fetch: stub `global.fetch` and `self.fetch` instead of `window.fetch` to verify it picks it up
- [ ] Test isomorphic by e.g. taking out `window`
- [ ] Clean up, squash, final checks, ask for review
- [ ] Open `urlBase` PR

ryanbrainard:
@passcod Love the checklist! Looking forward to it. Let me know if I can help with anything.

passcod:
Interesting consequence of this PR: I can try things out in a node REPL without the library complaining about missing fetch/Request objects. The isomorphic people are going to be happy. On a related note, does anyone know if it's possible for `expect` to show what exception it got, instead of just what it expected and didn't get? I get lots of failures like this:

```
  1) React connect .defaults should throw unless a number >= 0 is given as refreshInterval:
     Invariant Violation: refreshInterval must be positive or 0.
```

where I'd really like to see what _actually happened_ not just what _didn't_.

passcod:
I can't see a test for `force`'s behaviour in the original suite, so I'm not sure how to test that `force` is being overriden in defaults. Not sure it _makes sense_ for that to happen, but still.

passcod:
Here's a thought: should `.defaults()` also set `defaultMapPropsToRequestsToProps`?

README.md
```diff
@@ -341,6 +341,71 @@ userFetch.meta.response.headers.get('FOO')
...
+
+    if (mapping.url.match(/private/)) {
+      options.headers['X-Api-Token'] = getPrivateToken()
+    }
+
+    if (mapping.url.match(/listOfServers.json$/)) {
+      options.integrity = 'sha256-BpfBw7ivV8q2jLiT13fxDYAe2tJllusRSZ273h2nFSE='
+    }
+
+    return new Request(urlJoin(baseUrl, mapping.url), options)
```
nfcampos: This example suggests there's no use case for providing a default `Request` which is not covered by changing `buildRequest`, seeing as you could just `return new MyCustomRequest(...)` at the end of your custom `buildRequest`. Is there a use case for replacing `Request` without replacing `buildRequest`?
We at least need a warning preventing the user from supplying both custom `Request` and custom `buildRequest`, since a custom `Request` would only be used by the default `buildRequest`

README.md
```diff
@@ -16,6 +16,14 @@ npm install --save react-refetch
 
 This assumes that you’re using [npm](http://npmjs.com/) package manager with a module bundler like [Webpack](http://webpack.github.io) or [Browserify](http://browserify.org/) to consume [CommonJS modules](http://webpack.github.io/docs/commonjs.html).
 
+The following ES6 functions are required:
+
+- [`Object.assign`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)
+- [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)
+- [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/GlobalFetch) (This was previously included in builds, but isn't anymore.)
```
nfcampos: strictly speaking `fetch` isn't ES6, it's just a browser API, but it doesn't really matter

README.md
passcod: For your first point: yes, there is, when replacing `fetch` but not `buildRequest`.

README.md
passcod: Yeah, I can't find a way to rephrase that accurately. "Modern platform features" feels clunky and "ES6 functions and browser APIs" is too verbose. Idk.

README.md
passcod: For your second: I added it onto the list, but am wondering about the future. Given that it is possible that a future React Refetch may very well use `Request` outside of `buildRequest`, and that including both is relatively harmless, is it really useful to `console.warn()` the users versus just adding a note in the readme?

edoardo-bluframe:
Hey you guys - amazing work!

My React app is using react-refetch - it's awesome - and I'm making my app isomorphic this weekend... and I ran into the the wahtwg-fetch issue when I import the Routes from node!

Do you know when you're planning to merge the pull request?

passcod:
It should be finished tomorrow (so ~24 hours, it's late night here), and after that the review will take a little bit. I'd say it will be merged next week, tops? But I can't really predict those things.

edoardo-bluframe:
Awesome - that would be amazing

Is there a temporary solution you can think of so I can get the code to run?

Or is it stop the iso-morphing till the PR is merged?

passcod:
Isomorphic discussion should probably be done on #11. Note that this PR is very much _not_ focused on delivering a complete solution for isomorphic React Refetch. I have not and will not test more than what I have outlined in the checklist in this regard, and I strongly suspect that there will need to be further work to make it work completely in Node for the purpose of isomorphic rendering.

A temporary solution to get the code to run is to just use my branch, although it may break suddenly. I recommend you just wait.

passcod:
@ryanbrainard: Review please! :)

As mentioned above in https://github.com/heroku/react-refetch/pull/76#issuecomment-201687379 the only thing I'm not sure about is whether `force` and `comparison` make sense in defaults. I've added them anyway, but there are no tests for them. There is no test at all for `force` as of before this PR, and I can't figure out how to test for `comparison` when used in defaults.

At @nfcampos's suggestion https://github.com/heroku/react-refetch/pull/76#discussion_r57512221 I added a console info message if both `Request` and `buildRequest` are specified. I did not make it a warning as doing it anyway does not really *do* anything.

There are tests where I just wholesale copied source from previous tests or from connect.js source. I don't feel great about that, especially for the connect.js source material, but short of exporting some internal connect.js functions (the default internal `buildRequest` and `handleResponse`, notably), I don't really see a way around it.

nfcampos:
It looks like you didn't make it possible to set `fetch`, `handleResponse` and `buildRequest` at the individual mapping level, is that on purpose?

nfcampos:
Also, as long we're doing sanity checks on the arguments to `.defauts` might as well do it for all, then, catch, andThen, andCatch, force, refreshing are missing. What do you guys think?

passcod:
`force`, `refreshing`, and a few others are Booleans, but they're Booleanish-ly evaluated, so they can be falsy/truthy. Checks thus don't make any sense, given we can supply literally anything.

`then`, `catch`, and others take functions, sure, but they can also take *anything else* and that will be evaluated as "not a function" and thus will not be used.

Both these behaviours are what was already happening, so I just didn't change them.

nfcampos:
@passcod fair enough, but I don't think it's a great experience if someone erroneously supplies a non-function as eg. `andThen` and we silently ignore it instead of warning, since supplying a non-function to those keys is certainly a mistake.

passcod:
And yes, setting `fetch` etc at the mapping level is not possible. I don't believe fetch-in-mappings to be particularly intuitive, especially once you start adding fetch functions.

nfcampos:
but there are pretty valid use cases for them (at least some of them) at the mapping level, eg.
```js
connect(props => ({
  userFetch: 'api/some/json', // will use the default json-ready handleResponse
  dataFetch: {
    url: 'api/some/csv',
    handleResponse(resp) { return resp.text().then(...) }
  },
})

**Comment by passcod** [comment]
I know it's a bit of a shitty feeling, but this PR has taken literally 20 times more time than I originally planned for, and even though actually implementing that change would be 8 lines, _testing_ it would be a nightmare. For me. I think, anyway. I recognise there are some usecases for it. (You can also implement your example using a "global" `handleResponse` that switches based on the URL/Content-Type/whatever, so it's not like it's impossible to achieve this in the first place.) I'm just a bit tired of this. So, I'm asking for help? Here's a minimal patch for adding the feature:

```
diff --git a/src/components/connect.js b/src/components/connect.js
index b20bfd6..e15abd9 100644
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -207,0 +208 @@ function connect(mapPropsToRequestsToProps, defaults) {
+        buildRequest: defaults.buildRequest,
@@ -209,0 +211 @@ function connect(mapPropsToRequestsToProps, defaults) {
+        fetch: defaults.fetch,
@@ -210,0 +213 @@ function connect(mapPropsToRequestsToProps, defaults) {
+        handleResponse: defaults.handleResponse,
@@ -215,0 +219 @@ function connect(mapPropsToRequestsToProps, defaults) {
+        Request: defaults.Request,
@@ -224 +228 @@ function connect(mapPropsToRequestsToProps, defaults) {
-    return new defaults.Request(mapping.url, {
+    return new mapping.Request(mapping.url, {
@@ -325 +329 @@ function connect(mapPropsToRequestsToProps, defaults) {
-          const request = defaults.buildRequest(mapping)
+          const request = mapping.buildRequest(mapping)
@@ -329 +333 @@ function connect(mapPropsToRequestsToProps, defaults) {
-          const fetched = defaults.fetch(request)
+          const fetched = mapping.fetch(request)
@@ -332 +336 @@ function connect(mapPropsToRequestsToProps, defaults) {
-            return fetched.then(defaults.handleResponse).then(onFulfillment(meta), onRejection(meta))
+            return fetched.then(mapping.handleResponse).then(onFulfillment(meta), onRejection(meta))
```

File: README.md (hunk 8 above)
Comment (ryanbrainard) [inline README.md]: This was a peer dependency, but not actually included. I wonder if it would be good to add recommended polyfills for these. Don't block on this. We can always add something later.

**Comment by ryanbrainard** [comment]
> I know it's a bit of a shitty feeling, but this PR has taken literally 20 times more time than I originally planned for

I hear you. Let's get this merged sooner than later, and we can follow up with additional PRs as needed. (and a big thank you for your work on this!)

> Figure out what to do about `force` and `comparison` (does it even make sense to set them in defaults.

I definitely would use `comparison` as a default. In fact, I have a pattern that I use that is a data context container around other components that silently refreshes on changes to the path (i.e. { `comparison: location.pathname, refreshing: true }`). It's a really nice way of do local caching. I should probably write it up somewhere outside of this PR.

File: README.md [hunk 9]
```diff
@@ -341,6 +349,71 @@ userFetch.meta.response.headers.get('FOO')
 
 Do not attempt to read bodies directly from `meta.request` or `meta.response`. They are provided for metadata purposes only.
 
+## Setting defaults and hooking into internal processing
+
+It is possible to modify the various defaults used by React Refetch, as well as substitute in custom implementations of internal functions. A simple usecase would be to avoid repeating the same option for every fetch block:
```
ryanbrainard: use case (two words)

README.md
```diff
@@ -341,6 +349,71 @@ userFetch.meta.response.headers.get('FOO')
...
+const refetch = connect.defaults({
+  credentials: 'include'
+})
+
+refetch(props => ({
+  userFetch: `/users/${props.userId}`
+}))(Profile)
+```
+
+A more advanced usecase would be to replace the `buildRequest` internal function to, for example, modify headers on the fly based on the URL of the request, or using advanced [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request/Request) options:
```
ryanbrainard: use case (two words)

README.md
```diff
@@ -341,6 +349,71 @@ userFetch.meta.response.headers.get('FOO')
...
+import connect from './api-connector'
+connect(props => ({
+  userFetch: `/users/${props.userId}`,
+  serversFetch: `/listOfServers.json`
+}))(Profile)
+```
+
+### On changing the `fetch` and `Request` implementations
+
+Through this same API it is possible to change the internal `fetch` and `Request` implementations. This could be useful for a number of reasons, such as precise control over requests or customisation that is not possible with either `buildRequest` or `handleResponse`.
```
ryanbrainard: Replacing `handleResponse` isn't explained anywhere, is it?

README.md
```diff
@@ -341,6 +349,71 @@ userFetch.meta.response.headers.get('FOO')
...
+
+// ProfileComponent.js
+import connect from './api-connector'
+connect(props => ({
+  userFetch: `/users/${props.userId}`,
+  serversFetch: `/listOfServers.json`
+}))(Profile)
+```
+
+### On changing the `fetch` and `Request` implementations
```
ryanbrainard: this section doesn't really explain how to change the `fetch` and `Request` implementations, just warning that it is advanced

docs/api.md
```diff
@@ -13,11 +13,11 @@ Instead, it *returns* a new, connected component class, for you to use.
 
      - `url` *(String)*: Required. HTTP URL from which to fetch the data.
      - `method` *(String)*: HTTP method. Defaults to `GET`.
-     - `headers` *(Object)*: HTTP headers as simple key-value pairs. Defaults to `Accept` and `Content-Type` set to `application/json`.
-     - `credentials` *(String)*: Policy for credential to include with request. One of `omit`, `same-origin`, `include`. Defaults to `same-origin`. See [`Request.credentials`](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials) for details.
-     - `redirect` *(String)*: The redirect mode to use: `follow`, `error`, or `manual`. Defaults to `follow`.
+     - `headers` *(Object)*: HTTP headers as simple key-value pairs. Does _not_ support the [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers) object format. Defaults to `Accept` and `Content-Type` set to `application/json`.
```
ryanbrainard: thanks for doc'ing this. perhaps it should support `Headers` in the future, but I kept it as simple key-values because everything else was at that point.

docs/api.md
```diff
@@ -51,7 +48,40 @@ All the original static methods of the component are hoisted.
...
 ###### `getWrappedInstance(): ReactComponent`
 
-Returns the wrapped component instance. Only available if you pass `{ withRef: true }` as part of the `connect()`’s second `options` argument.
+Returns the wrapped component instance. Only available if you set `connect.defaults({ withRef: true })`.
+
+### `connect.defaults([newDefaults])`
+
+Returns a new `connect` which will have its defaults amended according to `newDefaults`.
+
+Calls to `.defaults()` can be chained, each latter call overriding defaults set in previous ones:
```
ryanbrainard: is it is full override or a merge?

ryanbrainard: similar question above about the use of the word "amended". not clear if that is overriding or merging the keys

docs/api.md
```diff
@@ -51,7 +48,40 @@ All the original static methods of the component are hoisted.
...
+```js
+const withRef = connect.defaults({ withRef: true })
+const withRefAndWithPost = withRef.defaults({ method: 'POST' })
+const onlyWithPost = withRefAndWithPost.defaults({ withRef: false })
+```
+
+#### Arguments
+
+* `[newDefaults = {}]` *(Object)*: An object with any of the following keys:
+     - `buildRequest(mapping): Request` *(Function)*: Takes a mapping and returns a `Request`. If setting this, make sure it interoperates with either the default `fetch` and `Request` or the `fetch` and `Request` you've provided. Defaults to the internal implementation, see the source for details.
```
ryanbrainard: will this be after coercion? in other words, will this function need to handle raw `url` strings or will `mapping` always be an object? I'm guess the latter, but I'm purposely reading the docs before looking at the source to have an outsider's mind.

docs/api.md
```diff
@@ -51,7 +48,40 @@ All the original static methods of the component are hoisted.
...
+     - `fetch(url|Request, options): Promise<Request>` *(Function)*: The function to use when performing requests. It should conform to the [`window.fetch` API](https://developer.mozilla.org/en-US/docs/Web/API/GlobalFetch/fetch). If setting this, make sure it interoperates with either the default `Request` or the `Request` you've provided. Defaults to the global `fetch`, if available, checking on `window`, `global`, and `self` in that order.
```
ryanbrainard: `window.fetch` or just `fetch`?

package.json
```diff
@@ -55,14 +55,14 @@
...
-    "webpack": "^1.11.0"
+    "webpack": "^1.11.0",
+    "whatwg-fetch": "*"
   },
   "dependencies": {
     "hoist-non-react-statics": "^1.0.3",
     "invariant": "^2.0.0"
   },
   "peerDependencies": {
-    "whatwg-fetch": "*",
```
ryanbrainard: so nice to see this go away! :boom:

src/components/connect.js
```diff
@@ -17,9 +27,127 @@ function getDisplayName(WrappedComponent) {
 // Helps track hot reloading.
 let nextVersion = 0
 
-export default function connect(mapPropsToRequestsToProps, options = {}) {
+function connectFactory(defaults = {}) {
+  function connectImpl(map, options = {}) {
+    /* eslint-disable no-console */
+    if (Object.getOwnPropertyNames(options).length > 0 && console && console.warn) {
+      console.warn('The options argument is deprecated in favor of `connect.defaults()` calls. In a future release, support will be removed.')
+    }
```
ryanbrainard: perhaps we can use https://www.npmjs.com/package/warning here. (don't block on this. we can do this later)

src/components/connect.js
```diff
@@ -70,30 +198,30 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
...
-        method: 'GET',
-        credentials: 'same-origin',
-        redirect: 'follow',
-        meta: {}
+        andCatch: defaults.andCatch,
+        andThen: defaults.andThen,
+        catch: defaults.catch,
+        credentials: defaults.credentials,
+        force: defaults.force,
+        meta: {},
```
ryanbrainard: any reason `meta` is not set in defaults? i realize it seems a bit silly as a default, but also seems strange to special case

src/components/connect.js
```diff
@@ -70,30 +198,30 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
 
   function assignDefaults(mapping, parent) {
     return Object.assign(
+      { comparison: defaults.comparison },
       parent ? {
```
ryanbrainard: note to self: re-think this whole `parent.comparison` thing. there's probably a better way. (future thing, not this PR)

src/components/connect.js
```diff
@@ -138,16 +266,16 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
...
         )
       }
 
       getWrappedInstance() {
-        invariant(withRef,
-          `To access the wrapped instance, you need to specify ` +
-          `{ withRef: true } as the fourth argument of the connect() call.`
+        invariant(defaults.withRef,
+          `To access the wrapped instance, you need to specify { withRef: true } ` +
+          `in the .defaults().`
```
ryanbrainard: no `the`

test/components/connect.spec.js
```diff
@@ -1177,5 +1199,502 @@ describe('React', () => {
...
+            `${name} must be a ${expectedType}. Instead received a ${type}.`
+          )
+        }
+
+        const checks = {
+          boolean: [ true, false ],
+          function: [ Date, () => {} ],
+          number: [ 1337, 1.37e25, 0.123456789, Math.PI, -42, 0, NaN, Infinity ],
+          object: [ null, [ 7.2, 'billion' ], new Date(), { obj: 'ect' } ],
+          string: [ 'string', 'chaîne', '細線', '😀🌂🔗💯🆗' ],
```
ryanbrainard: :heart:

docs/api.md
passcod: That's what I figured. `Headers` also doesn't have a simple way to merge, you'd have to iterate one to set values into the other, that's definitely more clunky than plain objects.

test/components/connect.spec.js
```diff
@@ -1177,5 +1199,502 @@ describe('React', () => {
     //  // But render is not because it did not make any actual changes
     //  expect(renderCalls).toBe(1)
     //})
+
+    context('.defaults', () => {
+      function regexpify(str) {
+        return new RegExp(str.replace(/([\[\]\{\}\(\)\.\-\,\+\?\*])/g, '\\$1'))
```
ryanbrainard: whoa, what's this little regex monster doing? ;)

docs/api.md
passcod: Yep, I had trouble phrasing this. Not sure if "merging" necessarily implies the overwriting. I added the example for that reason. I'll change to merge anyway in both places to keep it consistent.

ryanbrainard:
Looking really good overall. Mostly just small documentation things. I do think that everything should be available in both defaults and mappings, but that is something we can work on in the future.

I'm going to test this out on our internal app and let you know how it goes. If everything is good, I think we should pull this in very soon.

docs/api.md
passcod: This will be after coercion, yes.

docs/api.md
passcod: Um, not sure which instance of `fetch` you're talking about here.

If the link to MDN: Not sure. I wanted to keep the distinction with the method name and the React Refetch terminology.

If the "...global `fetch`, if available...", that's because it can either be `window.fetch`, `global.fetch`, or `self.fetch`. If that's unclear I can try to reword.

src/components/connect.js
passcod: Good idea, and it would avoid the clunky `console && console.warn` checking as well as the ESLint disables. I'll leave it for after review.

src/components/connect.js
passcod: Neither `meta` nor `value` are in defaults, mostly because I figured they were special cases already as they're only used for development/faking responses/ad-hoc custom responses in fetch block functions/etc.

src/components/connect.js
```diff
@@ -194,14 +322,14 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
         if (mapping.hasOwnProperty('value')) {
           return onFulfillment(meta)(mapping.value)
         } else {
-          const request = buildRequest(mapping)
+          const request = defaults.buildRequest(mapping)
           meta.request = request
           this.setAtomicState(prop, startedAt, mapping, initPS(meta))
 
-          const fetched = window.fetch(request)
+          const fetched = defaults.fetch(request)
```
ryanbrainard: This line fails on “Uncaught TypeError: Illegal invocation” in Chrome. See http://stackoverflow.com/a/9678166/129492 for why. Changing it to `const fetched = defaults.fetch.call(window, request)` fixes it, but this probably need to be done in a way that holds onto the reference to `window` and probably more places than just this.

src/components/connect.js
ryanbrainard: Fine to leave them out for now.

test/components/connect.spec.js
passcod: Escape characters that have special meaning in regexps. I'll add a comment.

passcod:
> In fact, I have a pattern that I use that is a data context container around other components that silently refreshes on changes to the path (i.e. `{ comparison: location.pathname, refreshing: true }`).

This sounds really cool!
I'll try to make a test for `comparison` in defaults.

src/components/connect.js
```diff
@@ -17,9 +27,127 @@ function getDisplayName(WrappedComponent) {
...
+    andCatch: undefined,
+    andThen: undefined,
+    buildRequest,
+    catch: undefined,
+    comparison: undefined,
+    credentials: 'same-origin',
+    fetch: top.fetch,
+    force: false,
+    handleResponse,
+    headers: {
```
ryanbrainard: what do you think about deep merging w/ `defaults.headers` here? in my test on a real app, i was just adding a custom header, but that overwrote these default JSON headers.

ryanbrainard:
Ugh, I realized that my use of `comparison` in defaults won't work because I don't have access to the props. I guess i could use the global `location.pathname`, but a better future fix would be to support `comparison` as a function.

ryanbrainard:
I tested it out on the internal app, and it worked great for the most part. The big problems were the `Uncaught TypeError: Illegal invocation` and lack of header merging. Let's get those fixed and bring this in.

cc: @jsullivan

src/components/connect.js
passcod: I wondered about that. I like the idea, but what if you actually want to remove these defaults? Hmm, I suppose someone wanting to remove those headers could just set them to `null` or `""` or something.

What about `.defaults()` chaining? Intuitively, we would also merge headers there, right?

src/components/connect.js
passcod: Hmm, tricky. I'll have a think.

passcod:
> I don't have access to the props

Yes, exactly, that's what I found when I tried to write tests. Supporting a function would be ideal, but probably not in this PR.

src/components/connect.js
ryanbrainard: True, but `Content-Type` and `Accept` are probably ones you should always set, so could be overridden. 

Related side note, this is kind of weird:

```
new Headers({foo: undefined}).get('foo')
"undefined"

new Headers({foo: null}).get('foo')
"null"
```

Perhaps if it is `null` or `undefined` we just remove them instead of letting it turned into a string like this.

src/components/connect.js
ryanbrainard: Oh, you might be able to use `bind` instead of `call`. I'm guessing that'll be easier to implement. See http://stackoverflow.com/a/34888425/129492

passcod:
That should be it for the code changes, I'll do the `handleResponse`, `fetch`, and `Request` documentation tomorrow.

passcod:
Rebased to master

README.md
passcod: Done below now.

ryanbrainard:
@passcod Good to go? You're probably itching to get this merger more than I am :)) Just tell me when and I'll press the big green button.

passcod:
Still a bit of docs to go, examples for `fetch`. I'm doing shopping now, but soon as get home :)

passcod:
Ok, I think that's it :)

ryanbrainard:
:boom: : :tada: :shipit:

ryanbrainard:
Release in [v1.0.0-beta.0](https://github.com/heroku/react-refetch/releases/tag/v1.0.0-beta.0)

ryanbrainard:
Wow, this ended up (directly and indirectly) closing 7 issues! Nice work @passcod!

PR #77, opened 2016-03-07
added other missing parens in the README

nfcampos:
There were two other snippets missing parens.


ryanbrainard:
Damn, that's what I get for hand writting these examples :-/ Thanks!

PR #80, opened 2016-03-09
(closes #79) Adds support for non-array iterables to PromiseState.all/race

### LINKED ISSUE #79: PromiseState.all/race should take iterables as documented
Broken out from https://github.com/heroku/react-refetch/issues/78#issuecomment-193502060

nfcampos:


nfcampos:
(btw, @ryanbrainard there's some problem with the tests when ran on watch mode, they always fail the 2nd time they run, ie. the first time they detect a change) It's always something saying window.Request is not defined (but it works when ran not on watch mode)

ryanbrainard:
Thanks! Hmm, not sure what's up with `watch` (i don't use it on this project)...

PR #84, opened 2016-03-14
Support falsey values

### LINKED ISSUE #83: support falsy values for value in mapping?
Hi,

The condition at https://github.com/heroku/react-refetch/blob/master/src/components/connect.js#L51 causes a request like
```js
connect(props => {
  return {
    blahFetch: {
      url: '/some/url'
      then: value => {
        if (value.some.key)
          return '/some/other/url'
        else
          return {value: null}
      }
    }
  }
})(MyComponent)
```
to be invalid. I understand the need to validate there is a `url` or `value`, however, how should I do a request like this? (ie. a request in which you poll a first url, you use the response of that to check whether something exists, if it exists you do the request you really want, otherwise that first request doesn't need a value)

Nuno

ryanbrainard:
Fixes #83 
@nfcampos look good to you?

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -48,8 +48,8 @@ export default function connect(mapPropsToRequestsToProps, options = {}) {
     }
 
     invariant(isPlainObject(mapping), 'Request for `%s` must be either a string or a plain object. Instead received %s', prop, mapping)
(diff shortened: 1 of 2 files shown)


nfcampos:
yep looks great!

ryanbrainard:
Released in [v0.7.1](https://github.com/heroku/react-refetch/releases/tag/v0.7.1)

PR #86, opened 2016-03-16
mapPropsToRequestsToProps is called with context as 2nd argument (closes #69)

### LINKED ISSUE #69: Add context to the parameters of the mapping function
Why don't we also pass context as a second parameter to the data-mapping function?
I looks like only props are passed:
https://github.com/heroku/react-refetch/blob/master/src/components/connect.js#L156

I wish I could also access the context like so:
```javascript
DataSourceEditModal = connect((props, context) => {

  // .... check context for router or other state in it. ....

  return { ...mapping... };
}
```

nfcampos:
@ryanbrainard any other test you can think is appropriate?
I'm testing
- context is indeed provided as the 2nd argument
- when context changes mapPropsToRequestsToProps is called with the new context

### DIFF
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -587,6 +587,65 @@ describe('React', () => {
       })
     })
 
+    it('should invoke mapPropsToRequestsToProps with context', () => {
+      let contextPassedIn = []
+
+      @connect((props, context) => {
+        contextPassedIn.push(context)
+        return {}
+      })
+      class InnerComponent extends Component {
+        render() {
+          return <div />
+        }
+      }
+      InnerComponent.contextTypes = {
+        foo: React.PropTypes.string
+      }
+
+      class OuterComponent extends Component {
+        constructor(props) {
+          super(props)
+          this.state = {
+            foo: 'bar'
+          }
+        }
+
+        getChildContext() {
+          return {
+            foo: this.state.foo
+          }
+        }
+
+        setFoo(foo) {
+          this.setState({ foo })
+        }
+
+        render() {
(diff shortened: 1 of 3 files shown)


test/components/connect.spec.js
```diff
@@ -587,6 +587,69 @@ describe('React', () => {
...
     })
 
+    it('should invoke mapPropsToRequestsToProps with context', () => {
+      let contextPassedIn = []
+
+      @connect((props, context) => {
+        contextPassedIn.push(context)
+        return {}
+      })
+      class WithProps extends Component {
```
ryanbrainard: maybe rename this to `InnerComponent` since it actually doesn't have any props

ryanbrainard:
Looks good! Can you update the API doc and maybe an example in the readme?

test/components/connect.spec.js
nfcampos: good point, I've renamed it.

nfcampos:
I'll update the docs.
As for the example, any example with context gets kind of verbose (defining all those contextTypes, etc.), which might be too much for the readme, no?

ryanbrainard:
> As for the example, any example with context gets kind of verbose (defining all those contextTypes, etc.), which might be too much for the readme, no?

Yeah, perhaps. I'll merge this now and then we can add more to the readme if it adds value.

Thanks for working on this @nfcampos !

nfcampos:
happy to help!

ryanbrainard:
Released in https://github.com/heroku/react-refetch/releases/tag/v0.8.0-beta.0

PR #89, opened 2016-03-27
added context example to README

nfcampos:
@ryanbrainard I finally got around to writing an example for context, what do you think, is this what you had in mind?

(related to https://github.com/heroku/react-refetch/pull/86)

I've also made two other changes in this PR (I wans't sure if I should open separate PRs for such small things):
- added the `curly` eslint rule as we discussed in a previous PR
- fixed `npm run test:cov` which wasn't working

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,7 +347,7 @@ You can also use `context` (in addition to props) to help define the requests yo
 It's available as the second argument to the function you pass to `connect`.
 Don't forget to define `contextTypes` on the component you're wrapping for context to be available.
 
-Some use case examples are, eg. setting a urlPrefix in a root component that all others use:
+Some use case examples are, eg. setting a URL prefix in a root component that all others use:
 ```js
 connect((props, context) => {
   return {
@@ -368,7 +368,7 @@ connect((props, context) => {
 })(Profile)
 ```
 
-This is a complete example of everything you need to use context with react-refetch:
+This is a complete example of everything you need to use context with React Refetch:
 ```jsx
 import React, { Component, PropTypes } from 'react'
 import { connect, PromiseState } from 'react-refetch'
--- a/README.md
+++ b/README.md
@@ -341,6 +341,83 @@ userFetch.meta.response.headers.get('FOO')
 
 Do not attempt to read bodies directly from `meta.request` or `meta.response`. They are provided for metadata purposes only.
 
+## Using context to define requests
+
+You can also use `context` (in addition to props) to help define the requests your components need.
+It's available as the second argument to the function you pass to `connect`.
+Don't forget to define `contextTypes` on the component you're wrapping for context to be available.
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -341,6 +341,83 @@ userFetch.meta.response.headers.get('FOO')
 
 Do not attempt to read bodies directly from `meta.request` or `meta.response`. They are provided for metadata purposes only.
 
+## Using context to define requests
+
+You can also use `context` (in addition to props) to help define the requests your components need.
+It's available as the second argument to the function you pass to `connect`.
+Don't forget to define `contextTypes` on the component you're wrapping for context to be available.
+
+Some use case examples are, eg. setting a urlPrefix in a root component that all others use:
```
ryanbrainard: s/urlPrefix/URL prefix/

README.md
```diff
@@ -341,6 +341,83 @@ userFetch.meta.response.headers.get('FOO')
...
+    userFetch: `/user/${props.params.userId}`,
+    likesFetch: {
+      url: `/users/${props.userId}/likes`,
+      refreshInterval: context.refreshInterval,
+    }
+  }
+})(Profile)
+```
+
+This is a complete example of everything you need to use context with react-refetch:
```
ryanbrainard: s/react-refetch/React Refetch/

ryanbrainard:
Looks good other than two simple little style nitpicks. 

Once #76 is pulled in, we might want to add some explanation explaining why you would set defaults in `connect.default` vs `context`, but we can do that later.

nfcampos:
I can add that note now if you want. The reason for doing it via context is basically so that it can change dynamically, right?

ryanbrainard:
> The reason for doing it via context is basically so that it can change dynamically, right?

Dynamically and/or more locally (depending on how global your defaults are). 

> I can add that note now if you want.

Let's hold off. I'm going to pull this in now, and push a release so that #76 will have its own beta.

nfcampos:
one other reason i can think of to use context:
an example like the default refreshTimeout I added to the readme in which you want to have a default for some mappings is impossible to do with `.defaults()`

ryanbrainard:
Yep, for sure. I like this use of context for defaults as you've shown.

Released: https://github.com/heroku/react-refetch/releases/tag/v0.8.0

loris:
@ryanbrainard @nfcampos Thanks for the example, I was wondering if there is a more DRYer way to make react-refetch access values in the context (in my case, authentication data such as API tokens) without explicitly doing this for every components using react-refetch (I have hundreds of them in a large codebase):

```
static contextTypes = {
  apiToken: PropTypes.string.isRequired,
}
```

I was thinking of wrapping my component with a "child-as-a-function" component (see example here: https://github.com/heroku/react-refetch/issues/140#issuecomment-245650583) instead of the HOC, what do you think?

Issue #92, opened 2016-03-28
PromiseState and any other kind of promised props


nfcampos:
So I've found a kind of sideways use for the ideas of PromiseState and connect (as a HOC that computes additional async props from your regular props). Components that need to compute props asynchronously derived from the other props need most of the ideas behind react-refetch (the props need to be represented as something that can be fulfilled, rejected, pending; it needs figure out whether to recompute when props change). In a way it's a lower-level primitive on top of which react-refetch is built. Do you think it belongs in react-refetch in any way?

I'm using it like this:
```js
export default withPromisedProps({
  table: {
    dependentProps: ['file'], // every time one of these props changes table is recomputed
    then: ({file}) => promisedFileReader(file).readAsText().then(csvParse)
  }
})(CsvTable)
```

Not sure if this is relevant to anyone, feel free to close this issue :)

ryanbrainard:
I need to think about this a bit more, but some initial thoughts:
 - can `dependentProps` be worked into `comparison` and/or `value` somehow? If possible, I'd like to have this based on identity requests. Perhaps `value` could take a `Promise` and do things like this.
 - I've considered breaking out `PromiseState` into a separate npm module. Perhaps that would help *if* we a separate module for this kind of thing.

nfcampos:
- `value` optionally taking a promise is a great idea, sounds like a very natural extension of the current API (and it looks like it'd be pretty trivial to implement, just change this line https://github.com/heroku/react-refetch/blob/master/src/components/connect.js#L345 to `return Promise.resolve(value).then(onFulfillment(meta), onRejection(meta))`
- as for `dependentProps`, it is kind of a specific implementation of a `comparison` function (called with (prevProps, nextProps) as arguments), if we had such a thing. But being able to specify comparison as an array of props to be shallowly compared actually sounds like something that would be useful for regular requests as well, what do you think? (btw, this api of dependentProps as an array of props to be compared was inspired by this https://github.com/acdlite/recompose/blob/master/src/packages/recompose/mapPropsOnChange.js)
- today I also considered whether PromiseState should be its own module, but honestly as long as you export it in react-refetch there's no practical difference at all, I had no trouble implementing this today importing PromiseState from react refetch

nfcampos:
#93 does the most significant part of this, so I'll close this issue

PR #93, opened 2016-03-29
Identity Request Promises

ryanbrainard:
If `value` is given a `Promise`, the `PromiseState` will be pending until the `value` or `reason` is settled; otherwise, the `PromiseState` will be resolved immediately.

This was suggested in #92 as a possible solution to @nfcampos desire to load files like fetches. It should also be compatible the expectation @neezer had in #63 that static data loads immediately. I think this is a win-win. Thoughts?

Note, I played a little bit w/ trying to unify fetch requests and identity requests and cut down on the nested conditionals, but the `meta` handling was getting in the way. If someone with more `Promise`-foo wants to take a stab at it, feel free :) Either way, I think the way it is now reads better, even though it is a bit verbose.


nfcampos:
I realised now that for this to be useful value has to be a function that returns a promise, instead of a promise, otherwise it will recompute every time anything about props changes, which seeing as these are potentially expensive operations (that's why they're async in the first place) is a deal breaker.

ryanbrainard:
@nfcampos Good point. Need to think on that.

nfcampos:
Assuming that any function that is assigned to `value` is a promise-returning function that we should call sounds too restrictive (I don't see a use case for using identity requests to inject a function, though, so maybe not?), another option is for this behaviour of promise identity requests to be under a new key `promise`. what do you think?

nfcampos:
This is definitely not something that I need in every component but I've now hit a second situation where I'd have used this, do you think this belongs in react-refetch @ryanbrainard ?

ryanbrainard:
@nfcampos is your previous comment still valid?

> I realised now that for this to be useful value has to be a function that returns a promise, instead of a promise

Are you saying your would like this as is or wrapped in a function?

nfcampos:
`value: somethingThatReturnsPromise()` works inside `then`, `andThen`, etc., in functional mappings. The only place where it might be problematic is directly as a top level request, but even then it's probably fine for most uses because we now only re-run the `mapPropsToRequestsToProps` function if the props changed. So I think it's fine as is, what do you think?

ryanbrainard:
Yeah, I think that makes sense, is simpler, and more consistent. Pulling this in.

PR #95, opened 2016-03-30
removed unused deepValue util, added 2 missing tests

nfcampos:
- removed unnecessary deepValue util
- ~~mapping.equals now short-circuits when it finds a key that is different~~ this was actually already true so I've reverted
- added tests for calling catch and andCatch mappings (which didn't exist)
- removed unnecessary branch in function with rest arguments (empty rest arguments are an empty array, which is truthy, so that branch was never taken


ryanbrainard:
LGTM! Thanks!

PR #96, opened 2016-03-30
type check mappings, allow all defaults at the mapping level, general housekeeping

nfcampos:
## todo

- [x] handleResponse moved to separate file in utils
- [x] buildRequest moved to separate file in utils
- [x] type checks moved to separate file in utils
- [x] tests that had copy-pasted handleResponse/buildRequest now import them
- [x] simplified defaults assignment
  - [x] fetch, handleResponse, buildRequest and Request can now be set at the mapping level
  - [x] every option can now be set at the individual mapping or defaults level
- [x] deprecated options argument to connect only sets withRef
- [x] added type checking for then, catch, andThen, andCatch
- [x] added tests checking type checking for then, catch, andThen, andCatch
- [x] added test checking that options at the mapping level take precedence over defaults
  - [x] tests that mapping level fetch, handleResponse, buildRequest, Request are used

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -136,22 +136,16 @@ function connect(mapPropsToRequestsToProps, defaults) {
   }
 
   defaults = Object.assign({
-    andCatch: undefined,
-    andThen: undefined,
     buildRequest,
-    catch: undefined,
-    comparison: undefined,
     credentials: 'same-origin',
     fetch: topFetch,
     force: false,
     handleResponse,
-    headers: {},
     method: 'GET',
     redirect: 'follow',
     refreshing: false,
     refreshInterval: 0,
     Request: topRequest,
-    then: undefined,
     withRef: false
   }, defaults)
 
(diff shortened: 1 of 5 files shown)


src/components/connect.js
```diff
@@ -17,34 +19,33 @@ function getDisplayName(WrappedComponent) {
...
-      'calls. In a future release, support will be removed.'
-    )
-
-    options.headers = mergeHeaders(options)
-    const finalOptions = Object.assign({}, defaults, options)
+    let finalOptions = defaults
+    if ('withRef' in options) {
+      warning(false,
+        'The options argument is deprecated in favor of `connect.defaults()`.' +
+        ' In a future release, support will be removed.'
```
passcod: IIRC, this warning is tested, so changing the wording may need a change in the test.

passcod:
One thing I wanted to do (but couldn't due to the one-file status of `connect.js`) when documenting #76 is that when, in the docs, there are places where it would be useful to refer to a particular place in the source, it would have been neat to actually link to the source.

Given that this PR is splitting everything up, that's now achievable :) For example, the `handleResponse` sample [here](https://github.com/heroku/react-refetch/blob/master/README.md#setting-defaults-and-hooking-into-internal-processing) would benefit greatly from a link to [the default implementation in the source](https://github.com/nfcampos/react-refetch/blob/per-mapping-fetch/src/utils/handleResponse.js).

src/components/connect.js
ryanbrainard: Still matches `/The options argument is deprecated/`

ryanbrainard:
This all looks really good! We might want to update the API doc to reflect some of these changes, but I don't want to block on that. Pulling this in now and feel free to send additional PRs for doc changes.

PR #100, opened 2016-04-07
fixes context (closes #99)

### LINKED ISSUE #99: context doesn't work
see gist at http://esnextb.in/?gist=03dcae315f27f7085504f3df0dfe6e46

basically it turns out that `hoist-non-react-statics` doesn't hoist `contextTypes`: https://github.com/mridgway/hoist-non-react-statics/blob/master/index.js#L9
which left me wondering how our tests didn't catch this, and it turns out that `context` "works" when you use `connect` as a decorator, because by the time you assign `contextTypes` to the component, that component is the wrapped version that results from `connect` (https://github.com/heroku/react-refetch/blob/master/test/components/connect.spec.js#L775).

Subtle bug...

@ryanbrainard this is pretty easy to fix (we just copy the contextTypes ourselves) but seeing as no one complained do we actually want to support context? it seems that no one ever used it, otherwise someone would have noticed it before me...

nfcampos:
closes #99

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -347,6 +347,10 @@ function connect(mapPropsToRequestsToProps, defaults) {
     RefetchConnect.displayName = `Refetch.connect(${getDisplayName(WrappedComponent)})`
     RefetchConnect.WrappedComponent = WrappedComponent
 
+    if (dependsOnContext && WrappedComponent.contextTypes) {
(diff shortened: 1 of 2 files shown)


ryanbrainard:
Awesome, thanks for fixing this!

ryanbrainard:

Issue #98, opened 2016-04-04
Update docs for allowing all defaults at the mapping level


ryanbrainard:
Docs for changes in https://github.com/heroku/react-refetch/pull/96. Particularly for:

 -  fetch, handleResponse, buildRequest and Request can now be set at the mapping level
  - every option can now be set at the individual mapping or defaults level

cc: @nfcampos  @passcod

PR #104, opened 2016-04-08
added .options() method for defining connect options

nfcampos:
- for now has only `withRef`

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,6 +1,6 @@
 ## API
 
-### `connect([mapPropsToRequestsToProps], [options])`
+### `connect([mapPropsToRequestsToProps])`
 
 Connects a React component to data from one or more URLs.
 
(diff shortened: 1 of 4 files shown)


ryanbrainard:
Awesome, thanks for doing this. Looks good, but API doc needs updating :)

nfcampos:
I've updated the API docs (also with the changes mentioned in #98). While I was at it I fixed two small things in the README

nfcampos:
@ryanbrainard I also added a test for an untested part of the `.defaults()` method — making sure it doesn't apply the defaults to the original `connect` it is called on. Also fixed the `comparison` test in the same manner as #101

docs/api.md
```diff
@@ -28,11 +28,17 @@ Instead, it *returns* a new, connected component class, for you to use.
      - `value` *(Any)*: Data to passthrough directly to `PromiseState` as an alternative to providing a URL. This is an advanced option used for static data and data transformations. Also consider setting `meta`.
      - `meta` *(Object)*: Metadata to passthrough directly to `PromiseState`. Keys `request`, `response`, and future keys may be overwritten.
 
-Requests specified as functions are not fetched immediately when props are received, but rather bound to the props and injected into the component to be called at a later time in response to user actions. Functions should be pure and return the same format as `mapPropsToRequestsToProps` itself. If a function maps a request to the same name as an existing prop, the prop will be overwritten. This is commonly used for taking some action that updates an existing `PromiseState`. Consider setting `refreshing: true` in such it situation.
+  The following keys may also be defined on an individual request (see their description below at `connect.defaults()`):
```
ryanbrainard: We might want to merge these all in one list. This is fine for now, but I might take another pass through the docs for #98.

ryanbrainard:
Awesome work, bringing this in.

docs/api.md
nfcampos: yeah I wasn't sure about making it two separate lists, the idea was that those 4 are kind of really not something you will usually use at the individual mapping level, but that's kind of a weak reason to separate them

ryanbrainard: yeah, i figured that was the case. i just pulled it in for now and will think about it. maybe how it is is fine.

ryanbrainard:

PR #106, opened 2016-04-09
Return existing PromiseState from PromiseState.resolve

### LINKED ISSUE #97: add static version of PromiseState#_mapFlatMapValue ?
That method is useful when building components that receive a prop that may or may not be a PromiseState.

eg.
```js
const Comp = ({table}) => {
  table = table instanceof PromiseState ? table : PromiseState.resolve(table)
  return (
    <div>
      {table.fulfilled && <span>{table.value.length}</span>}
    <div>
  )
}
```

(It should probably get a name like `PromiseState.ensure()`, mapFlatMapValue is kind of cryptic)

Sounds like a useful addition?

ryanbrainard:
Fixes #97 (and drops the now unneeded `#cast` from #105)

cc: @nfcampos

### DIFF
--- a/src/PromiseState.js
+++ b/src/PromiseState.js
@@ -16,8 +16,14 @@ export default class PromiseState {
     return ps
   }
 
-  // creates a PromiseState that is resolved with the given value
+  // creates a PromiseState that is resolved with the given value.
+  // if the given value is already a PromiseState,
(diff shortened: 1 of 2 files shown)


nfcampos:
Looks great :)

ryanbrainard:

PR #107, opened 2016-04-12
fixes handling of rejected fetches

nfcampos:
- fetch rejects rarely, but it does happen (eg. when the request times out), see https://fetch.spec.whatwg.org/#fetch-method for details
- fixed promise chaining in RefetchConnect#createPromise
- added tests for this behaviour
  - added another branch to our testing fetch stub (which rejects the promise)
  - added rejected fetch case to the 'should should props and promise state to the given component' test
  - tested that catch and andCatch handlers of a rejected request are called


ryanbrainard:
Thanks for fixing this!

ryanbrainard:
Released: https://github.com/heroku/react-refetch/releases/tag/v1.0.0-beta.4

PR #109, opened 2016-04-13
omit children in componentWillReceiveProps props comparison

nfcampos:
@ryanbrainard should I also omit `children` in the props object I call `this.refetchDataFromProps` with?

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -96,7 +96,8 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
   checkTypes(defaults)
 
   options = Object.assign({
-    withRef: false
+    withRef: false,
+    pure: true
   }, options)
 
   // Helps track hot reloading.
@@ -183,11 +184,17 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
 
       componentWillReceiveProps(nextProps, nextContext) {
         if (
+          !options.pure ||
           (dependsOnProps && !shallowEqual(omitChildren(this.props), omitChildren(nextProps))) ||
           (dependsOnContext && !shallowEqual(this.context, nextContext))
         ) {
           this.refetchDataFromProps(nextProps, nextContext)
         }
+      }
+
+      shouldComponentUpdate(nextProps, nextState) {
+        return !options.pure ||
+          this.state.data != nextState.data || !shallowEqual(this.props, nextProps)
       }
 
       componentWillUnmount() {
@@ -325,18 +332,22 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
 
           return {
             startedAts: Object.assign(
+              {},
               prevState.startedAts, {
                 [prop]: startedAt
               }),
             mappings: Object.assign(
+              {},
               prevState.mappings, {
                 [prop]: mapping
               }),
             data: Object.assign(
+              {},
               prevState.data, {
                 [prop]: datum
               }),
             refreshTimeouts: Object.assign(
+              {},
               prevState.refreshTimeouts, {
                 [prop]: refreshTimeout
               })
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -8,6 +8,7 @@ import PromiseState from '../PromiseState'
 import hoistStatics from 'hoist-non-react-statics'
 import invariant from 'invariant'
 import warning from 'warning'
+import omit from 'lodash/fp/omit'
 
 const defaultMapPropsToRequestsToProps = () => ({})
 
@@ -58,6 +59,8 @@ export default connectFactory({
     'Content-Type': 'application/json'
   }
 })
+
+const omitChildren = omit('children')
 
 function connect(mapPropsToRequestsToProps, defaults, options) {
   const finalMapPropsToRequestsToProps = mapPropsToRequestsToProps || defaultMapPropsToRequestsToProps
@@ -180,7 +183,7 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
 
       componentWillReceiveProps(nextProps, nextContext) {
         if (
-          (dependsOnProps && !shallowEqual(this.props, nextProps)) ||
+          (dependsOnProps && !shallowEqual(omitChildren(this.props), omitChildren(nextProps))) ||
           (dependsOnContext && !shallowEqual(this.context, nextContext))
         ) {
           this.refetchDataFromProps(nextProps, nextContext)
@@ -208,7 +211,7 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
       }
 
       refetchDataFromProps(props = this.props, context = this.context) {
-        this.refetchDataFromMappings(finalMapPropsToRequestsToProps(props, context) || {})
+        this.refetchDataFromMappings(finalMapPropsToRequestsToProps(omitChildren(props), context) || {})
       }
 
       refetchDataFromMappings(mappings) {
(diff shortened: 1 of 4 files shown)


ryanbrainard:
Yes, I think that makes sense. If not, there could be some subtle bugs where the someone is depending on `children`, but it doesn't actually update on change. I'd rather have it fail fast. I'm guessing you'd do that in https://github.com/nfcampos/react-refetch/blob/master/src/components/connect.js#L214?

nfcampos:
in two places i think: https://github.com/nfcampos/react-refetch/blob/master/src/components/connect.js#L213 and https://github.com/nfcampos/react-refetch/blob/master/src/components/connect.js#L189

nfcampos:
actually you're right, i'll do it just in the one you linked to

nfcampos:
done

nfcampos:
@ryanbrainard should I include the `shouldComponentUpdate` thing in this PR?

nfcampos:
(I also went ahead and updated the react devDependencies to v15)

ryanbrainard:
Sorry, which `shouldComponentUpdate` thing?

nfcampos:
similar optimisation but for preventing extraneous re-renders, the one we talked about here: https://github.com/heroku/react-refetch/pull/94#issuecomment-209108489

ryanbrainard:
Ah, yes, sure.

nfcampos:
should I make it optional, the component being `pure`, only re-rendering when props or state shallowly changed?

ryanbrainard:
Yes, we should default to `pure: true`, but allow apps to override with `pure: false`.

nfcampos:
should the pure option also control the componentWillReceiveProps optimisation? or in other words, setting `pure: false` only allows the render method of the wrapped component to be impure or also allows the `mapPropsToRequestsToProps` function to be impure?
(I'll do the API doc changes when we decide what the `pure` option controls)

src/components/connect.js
```diff
@@ -330,6 +341,7 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
                 [prop]: mapping
               }),
             data: Object.assign(
+              {},
```
nfcampos: should I make this change to the other state objects as well? I don't need to because I never compare them like I do `data`, but maybe it'd be more consistent?

test/components/connect.spec.js
```diff
@@ -1563,39 +1581,75 @@ describe('React', () => {
...
-    //
-    //  store.dispatch({ type: 'APPEND', body: 'a' })
-    //
-    //  // After store a change mapState has been called
-    //  expect(mapStateCalls).toBe(3)
-    //  // But render is not because it did not make any actual changes
-    //  expect(renderCalls).toBe(1)
-    //})
+    it('should re-render only when props or requests materially change (pure: true)', () => {
+      const Connected = connect(({ foo }) => foo ? { testFetch: `/resource/${foo}` } : {})(Passthrough)
```
nfcampos: this is implicitly assuming the default is `pure: true`, should it?

test/components/connect.spec.js
```diff
@@ -1118,7 +1139,6 @@ describe('React', () => {
           expect(fetchSpy.calls.length).toBe(2)
 
           // set BAR again, but will not be refetched
-          // TODO: no need to re-render here
```
nfcampos: I removed these because it's no longer true, there is a need to re-render there because the prop changes in a way that is not shallow equal

src/components/connect.js
```diff
@@ -192,7 +196,14 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
         this._unmounted = true
       }
 
+      shouldComponentUpdate(nextProps, nextState) {
+        return !options.pure ||
+          this.state.data != nextState.data ||
+          (dependsOnProps ? (this._propsDidChange || this.props.children != nextProps.children) : !shallowEqual(this.props, nextProps))
```
nfcampos: this would be simpler if I were to shallow compare the props again, is the complexity worth the performance gain from not calling `shallowEqual` again for the props?
(if the reason for `shouldComponentUpdate` being called is the props having changed, `componentWillReceiveProps` will have been called just before)

src/components/connect.js
```diff
@@ -180,7 +184,7 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
 
       componentWillReceiveProps(nextProps, nextContext) {
         if (
-          (dependsOnProps && !shallowEqual(this.props, nextProps)) ||
+          (dependsOnProps && (this._propsDidChange = !shallowEqual(omitChildren(this.props), omitChildren(nextProps)))) ||
```
ryanbrainard: perhaps I'm missing something, but where is `_propsDidChange` defined?

ryanbrainard: oh, i see, you're defining it here and using it below

nfcampos: exactly, should i change this to be less confusing?

src/components/connect.js
ryanbrainard: It does add some complexity (and did confuse me for a minute as you see above), but I guess just leave it in since it's already there and tested.

ryanbrainard: When I was experimenting before with `children`, I think that `this.props.children != nextProps.children` might always be `true`. I probably don't understand the intricacies of when this it changes well enough, but I remember it not being equal when nothing was changing. Also, just wondering about the intention here. It is correct to re-render a parent when the children change?

src/components/connect.js
ryanbrainard: sure, go ahead

test/components/connect.spec.js
ryanbrainard: Yes, but maybe add 'implicitly' into the test description

src/components/connect.js
nfcampos: So writing a `componentShouldUpdate` for components that have children is kind of a bummer, because the children will always be different on every update, even if they didn't change. The only time this is not so is when the component doesn't have children, in which `this.props.children == nextProps.children` is always true. Since `connected` components might or might not have `children` this condition is there for the times when the wrapped component doesn't have children.

(You have to re-render a parent when the children change, otherwise the new children would never appear on the screen)

nfcampos: should i add a comment explaining `this._propsDidChange` is defined in `componentWillReceiveProps` and will be used in `componentShouldUpdate`?

ryanbrainard:
> should the pure option also control the componentWillReceiveProps optimisation? 

Yes, I think so. I'd like to limit the options and if something is impure, let's consider the whole thing impure.

src/components/connect.js
ryanbrainard: Ah, makes sense. 

Thinking about this more, let's pull out the `this._propsDidChange` to reduce the complexity. It just feels like a bug waiting to happen.

nfcampos: alright, i'll take it out

nfcampos:
I've made the changes you mentioned, just missing the tests for `componentWillRecceiveProps` when pure is false and updating the API docs

ryanbrainard:
Looks good. I'll wait for the remaining bits and then merge.

src/components/connect.js
```diff
@@ -180,19 +184,26 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
...
+          this.state.data != nextState.data || !shallowEqual(this.props, nextProps)
+      }
+
       componentWillUnmount() {
         this.clearAllRefreshTimeouts()
         this._unmounted = true
       }
 
       render() {
+        this._propsDidChange = false
```
ryanbrainard: this can be removed now, right?

nfcampos: yep, forgot about that, ive removed it now

nfcampos:
@ryanbrainard I've added the missing tests and updated the API docs

ryanbrainard:
Thank you!

PR #111, opened 2016-04-15
calling functional mappings return a promise

### LINKED ISSUE #58: Easy way to register to for a one time promise callback?
Rendering different promise statuses (pending, setteled, etc) is easy (yay!). 
Performing a function **one-time** once a promise settles seems hard.
It's very common that I want to perform some logic when a promise is settled - for example:
- forms (load data into state): fetch form data, once available, setState({ data: value }).
- forms (post data, then do something): when submitting a form - on success, redirect to so other page (with react-router), or then - show success toaster.

if i do this:
```javascript
  // no good for me
  componentWillReceiveProps(nextProps) {
    nextProps.fetchAccountPs.then(account => this.setState({ account }));
  }
```
It gets called every time props change, even though the fetch promise does not change, not good. I only want it to get called once. when (ps.pending => ps.setteled). 
if the promise mapping changes and fetch occurs again, only then call the .then() again.

right now I need to keep track of when (ps.pending => ps.setteled) happens myself which is tedious.
would be really nice to be able to do
```javascript
connect({
  fetchAccountPs: {
     method: "GET",
     url: "/some/url",
     callback: "onAccountFetched"    // <==  suggestion
  }
)...
```
and the HOC would call my component exposed onAccountFetched(value) *once*, using a ref.

what do you think?
is there another existing way i'm missing?

nfcampos:
Promise is resolved or rejected with the value of all mappings that functional mapping returns.

What do you think @ryanbrainard ?

would (partially) address #58

edit: in trying to write an example use case for this I'm finding it hard to come up with something that can't be done in existing `then` and `catch` hooks in the mappings (with access to the `props` and `context`), so maybe what I want is just to be able to return `undefined` in those two hooks

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -223,20 +223,20 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
 
       refetchDataFromMappings(mappings) {
         mappings = coerceMappings(mappings)
-        Object.keys(mappings).forEach(prop => {
+        return Promise.all(Object.keys(mappings).map(prop => {
           const mapping = mappings[prop]
 
           if (Function.prototype.isPrototypeOf(mapping)) {
             this.setAtomicState(prop, new Date(), mapping, (...args) => {
-              this.refetchDataFromMappings(mapping(...args))
+              return this.refetchDataFromMappings(mapping(...args))
             })
             return
           }
 
           if (mapping.force || !mapping.equals(this.state.mappings[prop] || {})) {
-            this.refetchDatum(prop, mapping)
+            return this.refetchDatum(prop, mapping)
           }
-        })
+        }))
       }
 
       refetchDatum(prop, mapping) {
@@ -262,8 +262,7 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
           meta.request = request
           this.setAtomicState(prop, startedAt, mapping, initPS(meta))
 
-          const fetched = mapping.fetch(request)
-          return fetched
+          return mapping.fetch(request)
             .then(response => {
               meta.response = response
               return response
@@ -290,8 +289,7 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
             }
 
             if (mapping.then) {
-              this.refetchDatum(prop, coerceMapping(null, mapping.then(value, meta), mapping))
-              return
+              return this.refetchDatum(prop, coerceMapping(null, mapping.then(value, meta), mapping))
             }
 
             this.setAtomicState(prop, startedAt, mapping, PromiseState.resolve(value, meta), refreshTimeout, () => {
@@ -299,6 +297,8 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
                 this.refetchDataFromMappings(mapping.andThen(value, meta))
               }
             })
+
+            return value
           }
         }
       }
@@ -307,8 +307,7 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
         return (meta) => {
           return (reason) => {
             if (mapping.catch) {
-              this.refetchDatum(prop, coerceMapping(null, mapping.catch(reason, meta), mapping))
-              return
+              return this.refetchDatum(prop, coerceMapping(null, mapping.catch(reason, meta), mapping))
             }
 
             this.setAtomicState(prop, startedAt, mapping, PromiseState.reject(reason, meta), null, () => {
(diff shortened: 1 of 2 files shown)


test/components/connect.spec.js
```diff
@@ -332,12 +332,20 @@ describe('React', () => {
       expect(decorated.state.data.testFunc).toBeA('function')
       expect(decorated.state.data.deferredFetch).toEqual(undefined)
 
-      decorated.state.data.testFunc('A', 'B')
+      decorated.state.data.testFunc('A', 'B').then(value => expect(value).toEqual({ T: 't' }))
```
ryanbrainard: perhaps add something to this test to know that `then` was actually called

ryanbrainard: ...also, this assertion looks wrong. shouldn't it be something like `value.deferredFetch.value`?

ryanbrainard:
I like the way the returned `Promise` cleans things up internally, but it's a little strange how this is only helpful for functional mappings, and I fear that will expose too much of the internals. For #58, it seems like we need more universal and opaque hooks. I'd be ok with using the Promise internally, but not so sure about actually returning it to the app.

test/components/connect.spec.js
nfcampos: Ah, so the way I did this is the promise resolves not with the PromiseState but with just the value (it was less code this way, and I didn't see any use for the rest of the PromiseState, but now I'm thinking maybe `meta` would have been useful)

nfcampos:
Yeah I agree this doesn't look like an amazing solution. I actually did it partly because it was so easy, most of the necessary code was already in place.
I actually think the existing hooks are enough if we allow returning `undefined` for when you need to use them for side effects

ryanbrainard:
If we're on the fence on this, what about just omitting the final `return`? That way we pull in the refactorings to use `Promise`, but without changing the API. Think it is worth it?

> I actually think the existing hooks are enough if we allow returning undefined for when you need to use them for side effects

That is probably a good idea anyway.

src/components/connect.js
```diff
@@ -223,20 +223,20 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
 
       refetchDataFromMappings(mappings) {
         mappings = coerceMappings(mappings)
-        Object.keys(mappings).forEach(prop => {
+        return Promise.all(Object.keys(mappings).map(prop => {
           const mapping = mappings[prop]
 
           if (Function.prototype.isPrototypeOf(mapping)) {
             this.setAtomicState(prop, new Date(), mapping, (...args) => {
```
ryanbrainard: doesn't this need a `return` and inside of `setAtomicState` as well? (otherwise the `return` below isn't doing anything)

nfcampos: I think this one doesn't need a `return`because it's a functional mapping, there's no useful return value for a functional mapping (I guess), so I just placed the return below and it returns `undefined`

src/components/connect.js
```diff
@@ -307,15 +307,16 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
...
+              return this.refetchDatum(prop, coerceMapping(null, mapping.catch(reason, meta), mapping))
             }
 
             this.setAtomicState(prop, startedAt, mapping, PromiseState.reject(reason, meta), null, () => {
               if (mapping.andCatch) {
                 this.refetchDataFromMappings(mapping.andCatch(reason, meta))
               }
             })
+
+            throw reason
```
nfcampos: This line I'm not sure about, because up until now the promise created in `createPromise` was handled (ie. became a fulfilled promise) if it rejected. With this change, re-throwing the exception, we're making it an un-handled promise, which will emit the `unhandledRejection` event that browsers are starting to implement. I'm not sure it makes sense for react-refetch, as a library, to leave the promise unhandled like that, especially since we do handle it (set the correct state in the component etc.)

nfcampos:
If not for what I mentioned in this comment https://github.com/heroku/react-refetch/pull/111/files#r60853379 I'd agree with leaving this in without the final `return`. Because of that, and considering the limited utility, I'm on the fence. What do you think about the issue of leaving promises unhandled?

test/components/connect.spec.js
nfcampos: Good point about making sure `then` was actually called. if we decide to keep this i'll add that to the test

nfcampos:
one possible benefit of pulling this promise refactor in is that it allows us to register a general `onError` handler for all `react-refetch` asynchronous operations, not sure we need that though

loris:
+1

ryanbrainard:
Due to lack of recent activity on this PR, I'm closing it. If this is something still requested, please open a new PR.

PR #118, opened 2016-04-28
Delay json parsing

thebuilder:
Prevents SyntaxError being thrown, when trying to parse invalid json. Like the result of a 404 or 500 status.


ryanbrainard:
The problem with this is that it would then break JSON parsing for errors that are actually in JSON, which is really the how the error should be formatted. This handler should probably be improved to be more flexible for cases like this, but this change would not be backward compatible. If you want to use this implementation for your own app, you can set it as a custom handler like so:

```jsx
const customConnect = connect.defaults({
  handleResponse: function handleResponse(response) {
    ...
  }
})
```

Then use `customConnect` as you would `connect`. See https://github.com/heroku/react-refetch#setting-defaults-and-hooking-into-internal-processing for details.

Issue #119, opened 2016-05-01
setting a default `then` results in infinite requests


nfcampos:
We didn't realise when we added the capability to set default `then`, `andThen`, etc. that it results in infinite requests. Because the defaults also apply to the `then` request so this
```js
connect.defaults({then: '/other/url' })(() => ({
  abc: '/some/url'
})
```
becomes
```js
connect(() => ({
  abc: {
    url: '/some/url'
    then: {
      url: '/other/url'
      then: {
        url : '/other/url'
        ...
      }
    }
  }
})
```
First, we didn't realise this at the time because 2 tests that run before those for default `then` replace `window.fetch` and don't restore it back to the original mock. As such, the test requests never `resolve` and it seems to work... I'm fixing the tests, I'll open a PR soon. (done: see #120)

Second, I'm not sure how to actually fix this, do we have to special case `then` mappings so that they don't get the default `then`? How about `andThen` mappings, they also don't get the default `andThen`?

nfcampos:
Thinking about this a bit more it seems that default `andThen` and default `andCatch` are fine because the comparison prevents the requests from going forward. (if the default `andThen` has `force: true` it might fall into this bug as well)

jsullivan:
@nfcampos I think you're right about `andThen` and `andCatch`. I need to give this more thought because I haven't come up with a way to fix this that I'm happy with yet either.

nfcampos:
@jsullivan yeah I'm not sure how to fix this bug and still cleanly support setting a default for `then`. I don't see any way of doing so that doesn't treat `then` mappings as somehow special, which I'd like to avoid

ryanbrainard:
What about making it where `then` is only applied if it is different from the parent's `then`? I can't think of a good recursive case, but if there is one, apps could always re-create the function by using `Function.prototype.bind()` (kind of hacky, but maybe ok for such a corner case?). I'll poke around the code to see if this will work...

ryanbrainard:
Fixed by #123

PR #120, opened 2016-05-01
fixed window.fetch mocks leaking from one test to another

nfcampos:
- test are failing now because this uncovered an issue with setting a default `then`, see #119


nfcampos:
this contains important fixes for the tests (to make sure something like #119 doesn't happen again) so it should be pulled in as soon as we find a nice way of fixing #119

ryanbrainard:
Included in #123

PR #121, opened 2016-05-12
ignore then/catch mapping result if returned undefined

nfcampos:
this allows the use of `then` and `catch` for side effects. I chose to ignore the return value only if it is `undefined`, anything else will go through the same logic as before (eg. throwing if you return something other than a plain object, etc.)

partially addresses #58

remind me to update the docs before you merge this

PR #122, opened 2016-05-12
added refresing function to change value before request

### LINKED ISSUE #113: optimistic updates
I had the need for optimistic updates and came up with this
````js
connect(({params}) => ({
  project: `/project/${params.id}`,
  save: (project, name) => ({
    project: {
      value: {...project, name}
      andThen: () => ({
        project: {
          url: `/project/${params.id}`,
          method: 'PATCH',
          refreshing: true,
          body: {name}
        }
      })
    }
  })
}))(ProjectPage)
```
this works great, but it could be made much more convenient like this
```js
connect(({params}) => ({
  project: `/project/${params.id}`,
  save: (name) => ({
    project: {
      url: `/project/${params.id}`,
      method: 'PATCH',
      optimistic: true,
      refreshing: true,
      body: {name}
    }
  })
}))(ProjectPage)
```
What do you think @ryanbrainard ? Setting optimistic on a `PATCH` request would turn it into the first snippet, doing `Object.assign` of the current value of the request with the same name with the body of the optimistic `PATCH`. It would sort of be an extension of `refreshing`. (This is especially annoying to do by hand because it requires me to change the signature of the `save` function so that the current value is available to merge...)

nfcampos:
- if a mapping sets `refreshing` to a function then that mapping will be treated as `refreshing` *and* that function will be used to merge the existing value with anything you want (probably the body of the new request) *before* the request happens.
- this basically allows optimistic updates with either PATCH or PUT requests that replace a previous GET without introducing any new keyword for mappings

closes #113 

(if I haven't yet done so remind me to update the docs before you merge this)


jsullivan:
Hey @nfcampos this looks good to me. I'm going to go ahead and bring in this PR and #121 but did you want to update the docs first?

nfcampos:
@jsullivan I've updated the docs. I also made refreshing functions be ignored if it is set on a mapping that was never fulfilled before, I'd forgotten this the first time around.

jsullivan:
:thumbsup: merging.

ryanbrainard:
👍

PR #123, opened 2016-05-24
Fix infinite then

ryanbrainard:
This builds on https://github.com/heroku/react-refetch/pull/120 and suggests a fix for https://github.com/heroku/react-refetch/issues/119. I'm not sure I love the special casing going on here, but this does at least seem like a reasonable default to not inherit `then` and `andThen` from the `parent`. I did not do the same for `catch` and `andCatch` because I'm assuming those would be default error handlers. Please let me know what you think.

In testing this, I realized that `fetch`, `buildRequest`, and `handleResponse` were not inherited. It seems like they should be. Thoughts? Any others we should consider?

cc: @jsullivan @nfcampos

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -164,7 +164,14 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
         meta: {}
       },
       defaults,
-      parent ? { comparison: parent.comparison } : {},
+      parent ? {
+        fetch: parent.fetch,
+        buildRequest: parent.buildRequest,
+        handleResponse: parent.handleResponse,
+        comparison: parent.comparison,
+        then: undefined,
+        andThen: undefined
+      } : {},
       mapping,
       { headers }
     )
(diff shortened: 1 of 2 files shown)


nfcampos:
hi @ryanbrainard :)

so this would make it so that mappings that have a parent just do not inherit default `then` and `andThen`, right? I think this ends up being the most straightforward solution, so it's less likely it'll surprise someone.

src/components/connect.js
```diff
@@ -164,7 +164,14 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
         meta: {}
       },
       defaults,
-      parent ? { comparison: parent.comparison } : {},
+      parent ? {
+        fetch: parent.fetch,
+        buildRequest: parent.buildRequest,
+        handleResponse: parent.,
```
nfcampos: `handleResponse: parent.handleResponse,`

ryanbrainard: ugh, not sure how that happened... must have deleted it between running tests and committing. fixed

ryanbrainard:
Yes, that is what it would do. I had more complicated logic that involved comparing `defaults.then` and `parent.then`, but I realized you would probably never want to inherit these from `parent`. If someone really wants to use the same `then` in some kind of recursive scenario, they can always explicitly do so in the mapping itself. I also like this because while it is still special cased, at least it is like the existing things for mappings with a `parent`.

PR #124, opened 2016-05-25
fixes mocha watch mode

nfcampos:


test/components/connect.spec.js
```diff
@@ -203,8 +202,8 @@ describe('React', () => {
 
       const pending = TestUtils.findRenderedComponentWithType(container, Container)
       const startedAt = pending.state.startedAts.testFetch
-      expect(startedAt.getTime()).toBeLessThan(new Date().getTime())
```
nfcampos: I'm not sure why but when run in watch mode this test was failing because these two times were equal

test/setup.js
```diff
@@ -21,3 +15,13 @@ expect.extend({
     })
   }
 })
+
+if (!global.document) { // to make mocha's watch mode work
```
nfcampos: before the 2nd run in watch mode the jsdom globals would be recreated and the fetch polyfill would get confused

test/components/connect.spec.js
ryanbrainard: I've had this test fail locally with just `npm test` a few times like this. Were you seeing it consistently?

ryanbrainard: ah, i see the fix below. that should do it

ryanbrainard:
Thanks!

ryanbrainard:
Since this is a test-only change, not planning on doing a special release for it.

test/components/connect.spec.js
nfcampos: yeah in watch mode the 2nd time the test ran it always failed there

PR #125, opened 2016-05-25
compare mapping against parent if it exists

### LINKED ISSUE #102: request shallow comparison with then() should compare the original mapping?
This is similar to #53 but not exactly the same because I think this `parent` comparison should be true also when you don't provide `comparison`
eg.
```js
@connect(({ foo }) => ({
  testFetch: {
    url: `/resource/${foo}`,
    then(value) {
      return { value }
    }
  }
}))
class WithProps extends Component {
  ...
}
```
currently the comparison of requests compares `{ value }` that resulted out of the `then` with `{ url: `/resource/${foo}` }` of then new original request, this makes the comparison logic all but useless when `then` is used. It seems to me that the expected behaviour would be for the new original mapping to be compared with the previous original mapping (not the one resulting from then). Do you agree @ryanbrainard?

nfcampos:
- basically just compares new mappings against the parent (if it exists). only goes one level deep so if you have 
```js
{
  foo: {
    url: '/foo',
    then: () => ({
      url: '/then',
      then: () => ({
        url: '/thenthen'
      })
    })
  }
}
```
it'll still be broken

- missing a test for this, don't merge yet

closes #102 

what do you think of this solution @ryanbrainard ?

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -144,7 +144,7 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
       return [ 'value', 'url', 'method', 'headers', 'body' ].every((c) => {
         return shallowEqual(this[c], that[c])
       })
-    }.bind(mapping)
+    }.bind(parent || mapping)
 
     return mapping
   }
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -132,6 +132,10 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
 
     checkTypes(mapping)
 
+    if (parent) {
+      mapping.parent = parent.parent || parent
+    }
+
     mapping = assignDefaults(mapping, parent)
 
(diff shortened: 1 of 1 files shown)


ryanbrainard:
Hmm...yeah, I think that'll work. Might need to think on it. Also, maybe `parent` should always be the top-level mapping as to work in cases with multi-level chains?

nfcampos:
Yeah I agree `parent` should always be the top level mapping

nfcampos:
the cleanest way to do that is to save a reference to `parent` (if it exists) in each mapping, inside `coerceMapping`, any reason not to do that?
like this:
screenshot 2016-05-25 19 40 31

ryanbrainard:
Seems like that should work

ryanbrainard:
Thanks!

PR #127, opened 2016-05-27
added test for #125

nfcampos:
turns out I had forgotten a test for #125 
and apparently it was needed, it is the `that` mapping in the `equals` function that needs to be always the parent (because the `that` mapping is the older one, `this` is the new mapping)


ryanbrainard:
Ah, good call! Thanks

PR #128, opened 2016-06-01
allow then, andThen, catch, andCatch to be undefined

nfcampos:
- extended `typecheck` function to support more than on allowed type
- extended `typecheckCheck` test function to support more than one allowed type

### DIFF
--- a/src/utils/checkTypes.js
+++ b/src/utils/checkTypes.js
@@ -1,10 +1,12 @@
 import invariant from 'invariant'
 import isPlainObject from './isPlainObject'
 
-function typecheck(type, name, obj) {
+function typecheck(types, name, obj) {
   invariant(
-    typeof obj === type,
-    `${name} must be a ${type}. Instead received a %s.`,
+    Array.isArray(types)
+      ? types.some(t => typeof obj === t)
+      : typeof obj === types,
+    `${name} must be a ${types}. Instead received a %s.`,
     typeof obj
   )
 }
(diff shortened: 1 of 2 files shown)


src/utils/checkTypes.js
```diff
@@ -1,10 +1,12 @@
...
 
-function typecheck(type, name, obj) {
+function typecheck(types, name, obj) {
   invariant(
-    typeof obj === type,
-    `${name} must be a ${type}. Instead received a %s.`,
+    Array.isArray(types)
+      ? types.some(t => typeof obj === t)
+      : typeof obj === types,
+    `${name} must be a ${types}. Instead received a %s.`,
```
ryanbrainard: This message will be weird if `${types}` is an array. Probably need to do some `join`ing.  Also, it might be easier to convert `types` to an `Array` before the `invariant` to avoid having two special cases in here.

nfcampos: actually I tested it and an array is coerced to a string correctly, eg.
```js
arr = ['function', 'undefined']
str = `then must be a ${arr}`
```
`str` will be `then must be a function, undefined` which is pretty clear but when types is an array I can replace `a` with `one of` so that it becomes `then must be one of function, undefined`, should I do that?

ryanbrainard: yeah, it would probably read better that way. (Of course this really isn't super important since it is only something shown in the dev console...)

nfcampos: done

ryanbrainard:
Looks like the test needs to be updated.

nfcampos:
fixed now

ryanbrainard:
Thanks!

nfcampos:
@ryanbrainard could you cut a release with this?

ryanbrainard:
@nfcampos oh, sure, sorry I didn't already

ryanbrainard:
@nfcampos Released https://github.com/heroku/react-refetch/releases/tag/v1.0.0-beta.8

nfcampos:
@ryanbrainard Thanks!

PR #130, opened 2016-06-15
Fix cachingFetch example in doc

loris:
cachingFetch example is not working for 2 reasons:

- `new Date() + 60000` won't work, date needs to be casted to timestamp first otherwise it returns something like `Wed Jun 15 2016 16:25:57 GMT+0200 (CEST)60000`
- the returned value has to be a promise


ryanbrainard:
Thanks!

PR #132, opened 2016-06-20
Adds .babelrc to .npmignore

oblador:
I'm using React Native which is not precompiled (nor most of its third party libraries) I need to apply babel to npm dependencies and since babel tries to use configs for each package if present, the presence of it here breaks my setup.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,3 @@
 src
 examples
+.babelrc


ryanbrainard:
What do you think about doing like https://github.com/reactjs/react-redux/pull/213 instead?

ryanbrainard:
I'm not too opinionated about it, but since this was originally forked from react-redux (and sometimes used along side it), I like it keep things like this more or less the same.

oblador:
So, do you want a new PR?

ryanbrainard:
That would be great :)

oblador:
Aight, opened a new PR, closing this one.

PR #134, opened 2016-06-23
Remove redundant instanceof check

apalm:


ryanbrainard:
Good catch. I think this got missed in a refactoring. Looks like `_mapFlatMapValue` could be completely inlined now.

ryanbrainard:
Thanks!

PR #139, opened 2016-07-23
#58 Added wrappedInstance to the meta object

eyalw:
Added wrappedInstance to the meta object for side effects during then() and andThen().

the wrappedInstance is called just "component" in the context of then() because I thought it would be a simpler name yet clear enough.


eyalw:
fixes #58

src/components/connect.js
```diff
@@ -279,6 +279,7 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
         } else {
           const request = mapping.buildRequest(mapping)
           meta.request = request
+          meta.component = this.refs.wrappedInstance
```
ryanbrainard: this should probably be done right after line 267 (or something like `const meta = Object.assign(meta, {component: this.refs.wrappedInstance})` so that identity requests and regular fetches (i.e. the branches of the `if/else`) are treated the same in this regard.

ryanbrainard:
Thanks for opening this. Would you also be able to add a test (there's probably an existing one that can be easily modified), add it to the API doc, and perhaps add an example?

docs/api.md
```diff
@@ -22,6 +22,7 @@ Instead, it *returns* a new, connected component class, for you to use.
      - `force` *(Boolean)*: Forces the data to be always fetched when new props are received. Takes precedence over `comparison`.
      - `comparison` *(Any)*: Custom value for comparing this request and the previous request when the props change. If the `comparison` values are *not* strictly equal, the data will be fetched again. In general, it is preferred to rely on the default that compares material changes to the request (i.e. URL, headers, body, etc); however, this is helpful in cases where the request should or should not be fetched again based on some other value. If `force` is true, `comparison` is not considered.
      - `then(value, meta): request` *(Function)*: returns a request to fetch after fulfillment of this request and replaces this request. Takes the `value` and `meta` of this request as arguments. Return `undefined` in `then` to do side-effects after a successful request, leaving the request as is.
+      - `meta` is an object with keys `{ request, response, component, ...<your meta>}` where `component` is equal to the wrappedInstance. You can use it to create side effects on promise fulfillment. e.g. `then(value, meta) { meta.component.onDataLoaded(value); }`
```
nfcampos: only if you passed `withRef: true` to options, otherwise `component` will be `undefined`

ryanbrainard: Good point. I'll add this note after merging.

ryanbrainard:
Thanks @eyalw !

docs/api.md
ryanbrainard: Good point. I'll add this note after merging.

ryanbrainard:
I was just doing a quick test of this in a real app and got:

```
Invariant Violation: addComponentAsRefTo(...): Only a ReactOwner can have refs. You might be adding a ref to a component that was not created inside a component's `render` method, or you have multiple copies of React loaded
```

I haven't taken the time to dig in, but just wanted to check with @eyalw if you had this happen when using `connect.options({ withRef: true })`

eyalw:
I didn't, an odd message. Do you have a short code that duplicated the
warning?

eyalw:
@ryanbrainard are you releasing a new beta with this code soon?

ryanbrainard:
@eyalw Oh, yes, sorry, I can cut a release

ryanbrainard:
Released: https://github.com/heroku/react-refetch/releases/tag/v1.0.0-beta.9

PR #150, opened 2016-09-22
Fix 'PromiseState.refresh should return a new PromiseState'

### LINKED ISSUE #149: PromiseState.refresh should return a new PromiseState
PromiseState.refresh should return a new PromiseState because this way a `shallowEqual` comparison of the props in the wrapped component erroneously returns `true` even though the PromiseState changed

ryanbrainard:
Fixes #149 

cc: @nfcampos

### DIFF
--- a/src/PromiseState.js
+++ b/src/PromiseState.js
@@ -11,9 +11,17 @@ export default class PromiseState {
   // creates as PromiseState that is refreshing
   // can be called without a previous PromiseState and will be both pending and refreshing
   static refresh(previous, meta) {
-    const ps = previous || PromiseState.create(meta)
-    ps.refreshing = true
-    return ps
+    const p = previous || PromiseState.create(meta)
+
+    return new PromiseState({
+      pending: p.pending,
+      refreshing: true,
+      fulfilled: p.fulfilled,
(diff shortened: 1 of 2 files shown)


src/PromiseState.js
```diff
@@ -11,9 +11,17 @@ export default class PromiseState {
   // creates as PromiseState that is refreshing
   // can be called without a previous PromiseState and will be both pending and refreshing
   static refresh(previous, meta) {
```
ryanbrainard: I didn't address it here, but was noticing how it's a little weird that `meta` here is only used if `previous` is undefined; although, it does act like `resolve(value, meta)` though. I believe I did that this to simplify some of the logic on the refetch size, but it makes this API a little ambiguous. Don't think it needs to be addressed now, but just calling it out for the future.

nfcampos: Yeah that's a little weird

nfcampos:
This looks great! 👍

ryanbrainard:
Released: https://github.com/heroku/react-refetch/releases/tag/v1.0.0-beta.10

nfcampos:
Thanks!

PR #155, opened 2016-12-17
fix wrap-multilines eslint rule, closes #152

### LINKED ISSUE #152: eslint: wrap-multilines is depreciated
wrap-multilines in eslint has been depreciated, has changed to jsx-wrap-multilines

ianstormtaylor:


ryanbrainard:
Travis is failing on this rule. Are you sure its correct?

ianstormtaylor:
Ah no, I forgot that it would require `eslint-plugin-react` to be upgraded, my bad. I kinda just fired this off quickly because I saw https://github.com/heroku/react-refetch/issues/152 was open and I figured why not open a quick fix. But now that I look at it again, I'm not exactly sure why that issue is open, because the `eslint-plugin-react` is pegged so it should never be a problem 😳

Issue #166, opened 2017-02-20
pollingInterval not cancelled if request is outstanding


arbesfeld:
I have built a decorator like this:

```
export default ({ LoadingView = undefined, refreshInterval = 0 } = {}) =>
  WrappedComponent => ({ params }) => ({
    fetchURL: {
      url: `api/`,
      refreshInterval,
    },
  }(({
  fetchURL,
  ...rest
}) => {
    function onFulfillment(data) {
      return <WrappedComponent {...rest} data={data} />;
    }

    return (
    <PromiseStateContainer
      ps={fetchURL}
      onFulfillment={onFulfillment}
      onPending={LoadingView}
    />
    );
  }));
```

And using it like this:

```
import decorator from '../decorator';

@decorator({ pollingInterval: 2000 })
class MyClass extends Component {}
```

However, if I unmount the component while a poll is pending, then the polling continues indefinitely.

arbesfeld:
@ryanbrainard - this appears to happen whenever the first url doesn't "resolve", and then the component is unmounted

ryanbrainard:
@arbesfeld Thanks for the report. I haven't tried reproducing yet, but sounds plausible in the non-resolving case. Marking as a bug.

TacticalCoding:
I have noticed this behavior as well. I'm using react-router, and if I navigate to a new page before an API request has returned from react-refetch, the polling is not cancelled from the previous page we were on and I see requests for both pages.

ryanbrainard:
The fix from @TacticalCoding has been pre-released in [v1.0.1-0](https://github.com/heroku/react-refetch/releases/tag/v1.0.1-0). Please try it out! Will cut a patch release if all goes well.

3ll3d00d:
This fixes it in my app. When will you publish an official release?

PR #169, opened 2017-04-18
Don't setTimeout if component is unmounted.

TacticalCoding:
Fixes https://github.com/heroku/react-refetch/issues/166

This happens in the case where the previous fetch hasn't resolved yet
and the component is unmounted. In that case it keeps refreshing the
unmounted component forever.

Also includes fix for unrelated non-deterministic test that occurs because
the test assumes that two different lines in the test will take place more
than one millisecond apart, which is not always true.

### DIFF
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -203,7 +203,7 @@ describe('React', () => {
       const pending = TestUtils.findRenderedComponentWithType(container, Container)
       const startedAt = pending.state.startedAts.testFetch
       setImmediate(() => {
-        expect(startedAt.getTime()).toBeLessThan(new Date().getTime())
+        expect(startedAt.getTime()).toBeLessThanOrEqualTo(new Date().getTime())
         const fulfilled = TestUtils.findRenderedComponentWithType(container, Container)
         expect(fulfilled.state.startedAts.testFetch).toEqual(startedAt)
         done()
@@ -801,7 +801,7 @@ describe('React', () => {
     })
 
     it('should refresh when refreshInterval is provided', (done) => {
-      const interval = 100000 // set sufficently far out to not happen during test
+      const interval = 100000 // set sufficiently far out to not happen during test
 
       @connect(() => ({ testFetch: { url: `/example`, refreshInterval: interval } }))
       class Container extends Component {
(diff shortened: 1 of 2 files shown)


test/components/connect.spec.js
```diff
@@ -856,6 +856,77 @@ describe('React', () => {
...
+
+        // Cancel scheduled refresh
+        clearTimeout(refreshTimeout)
+
+        outerComponent.setRender(false)
+
+        // force the refresh to happen now after it has been unmounted
+        after()
+
+        const spy = expect.spyOn(window, 'setTimeout')
```
ryanbrainard: shouldn't this happen before `after()`? I'm also not really clear what's going on there with the `spy.destroy()` below..

test/components/connect.spec.js
```diff
@@ -856,6 +856,77 @@ describe('React', () => {
...
+        expect(refreshTimeout).toBeTruthy()
+        const after = refreshTimeout._onTimeout
+
+        // Cancel scheduled refresh
+        clearTimeout(refreshTimeout)
+
+        outerComponent.setRender(false)
+
+        // force the refresh to happen now after it has been unmounted
+        after()
```
ryanbrainard: can you explain what this is doing? it appears that it's just forcing the call the the `refreshTimeout._onTimeout`, but doesn't seem like that would actually test the code you added.

test/components/connect.spec.js
TacticalCoding: Hmm.. You're right that it is strange that it is after but I was following the `should not call setState if component is unmounted'` test which does it similarly. But it seems to be working like this because it fails without my change and passes with it.

test/components/connect.spec.js
TacticalCoding: I'm forcing the refreshTimeout._onTimeout after the component has been unmounted. If `_onTimeout` is called without the `setRender(false)`, the `setTimeout` gets called and the refresh continues indefinitely even though the component has unmounted.

ryanbrainard: makes sense

test/components/connect.spec.js
ryanbrainard: I just tested this (and the other test) with the `spy` up before `after()` and without the `spy.destroy()` and it worked fine. The reason it didn't really matter is that the `setImmediate` is what triggers the change by yielding to the promise. I'll merge this as is and then fix both of the tests.

ryanbrainard:
Thanks for the fix!

ryanbrainard:
Pre-released in [v1.0.1-0](https://github.com/heroku/react-refetch/releases/tag/v1.0.1-0).

Issue #171, opened 2017-05-02
Modifying headers in `buildRequest` causes fetch on every page resize


aaronschwartz:
I have a situation where I want to send a token back in the headers with every request. That token is subject to change throughout the life of the app. I want to put it in my helper for setting defaults so that I don't have to worry about it throughout my code.

If I do:
```
connect.defaults({
  headers: {
    "token": localStorage.getItem("token")
  }
})
```
then the token is not updated if it changes because it is only calculated once.

If I add it in `buildRequest`:
```
connect.defaults({
  buildRequest: (mapping) => {
    const options = {
      ...
      headers: mapping.headers
    }

    options.headers["token"] = localStorage.getItem("token");

    return new Request(mapping.url, options);
  }
})
```
then my requests are correct and the token is updated on every call, but my issue is that every time I resize my page it fires a component update, and when it does a header comparison it decides that it has changed (even if the token has not changed) and causes many fetches while resizing.

It looks like it is comparing the header before the `buildRequest` without the token and the header after the `buildRequest` with the token during it's equality check.

It feels like it should save the `[ 'value', 'url', 'method', 'headers', 'body' ]` after the buildRequest to use for comparison purposes to see if the request has changed.

Is there a better way to do what I'm looking for?

ryanbrainard:
Is this fixed by #172?

aaronschwartz:
Yep it should be. Thanks.

PR #172, opened 2017-05-08
Allow headers to be specified as functions.

aaronschwartz:
I couldn't find a workaround for #171 without making a library change. This allows you to set a dynamic header value while setting up defaults. I considered making the entire `headers` object a function but that required many more changes that I didn't want to mess with since there are various places that it is treated specifically as an object.

E.g:
```
connect.defaults({
  headers: {
    "token": () => localStorage.getItem("token")
  }
})
```

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -161,7 +161,9 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
     for (let key in rawHeaders) {
       // Discard headers with falsy values
       if (rawHeaders.hasOwnProperty(key) && rawHeaders[key]) {
-        headers[key] = rawHeaders[key]
+        // Get the value now if the header is specified as a function
+        const headerValue = typeof rawHeaders[key] == 'function' ? rawHeaders[key]() : rawHeaders[key]
+        headers[key] = headerValue
       }
     }
 
--- a/test/components/connect.spec.js
+++ b/test/components/connect.spec.js
@@ -255,6 +255,31 @@ describe('React', () => {
       expect(decorated.state.mappings.testFetch.equals).toBeA('function')
     })
 
+    it('should allow header values to be specified with functions', () => {
+      @connect(() => ({ testFetch: {
+        url: '/example',
+        method: 'POST',
+        headers: {
+          'Content-Type': 'overwrite-default',
+          'X-Foo': () => 'test-function'
+        }
+      } }))
+      class Container extends Component {
+        render() {
+          return <Passthrough {...this.props} />
+        }
+      }
+
+      const container = TestUtils.renderIntoDocument(
+        <Container />
+      )
+
+      let decorated = TestUtils.findRenderedComponentWithType(container, Container)
+      expect(Object.keys(decorated.state.mappings.testFetch).length).toEqual(14)
+      expect(decorated.state.mappings.testFetch.method).toEqual('POST')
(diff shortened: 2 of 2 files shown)


ryanbrainard:
Looks like a reasonable change, but looks like there's a test failure. Also, any thoughts on if this function should take any arguments? We can always add that later if a need arises, but just thinking about it.

nfcampos:
why not use a getter?
```js
connect.defaults({
  headers: {
    get token() { return localStorage.getItem("token") }
  }
})
```

aaronschwartz:
I thought of using the getter as well but the object is spread into another object early on (at which point the getter is evaluated), but only the getter's value is passed on so it never gets evaluated again:
https://github.com/heroku/react-refetch/blob/master/src/components/connect.js#L45

I'll fix the Lint warning.

nfcampos:
Fair enough, would it be better then to change the code to allow getters to be used, eg. using `Object.getOwnPropertyDescriptors` and `Object.defineProperties` instead of `Object.assign`?

aaronschwartz:
I don't really mind either way as far as API goes. There are a couple of places where the headers are object.assign'd and merged with overrides. I was trying to do the least amount of changes possible so I didn't break anything. I'm not too familiar with getOwnPropertyDescriptors and defineProperties.

nfcampos:
I tend to prefer to use language features in place of new APIs but don't mind too much either way.
Anyway, it's up to @ryanbrainard i'm just offering my 2 cents as someone very familiar with the library

ryanbrainard:
@nfcampos Can you provide an example of what you're suggesting. If it's possible without changes to the lib, that sounds better, but I'm not totally following your suggestion.

ryanbrainard:
@nfcampos Oh, now I see you did provide an example above in https://github.com/heroku/react-refetch/pull/172#issuecomment-300468745. 

Given the amount of `Object.assign` sprinkled in this lib, I'm a little hesitant to change how a lot of that works, but that's probably mostly my own unfamiliarity with the ins and outs of getters, `getOwnPropertyDescriptors`, and `defineProperties`. How much work do you see it being to make this change and how risky would it be for backward compatibility?

Given that other areas of this library support the raw value or function pattern (like the prop mappings themselves), which probably could have also been done with getters, I'm leaning toward merging this PR as is because it will be familiar. Then later, we could *consider* also supporting getters, but doing it everywhere, not just headers. Thoughts?

nfcampos:
@ryanbrainard I agree, this change should probably apply everywhere, not just to headers. And I'm afraid it should be a major version bump, because it does possibly change what values we use for properties in the mappers — ie. for users that were using getters we'd previously always be using the value the getter returned when we internally use `Object.assign`, and now we'd only invoke the getter once the value was actually used in a request.

The change would basically be replacing all usage of `Object.assign` with something like 
```js
function assign(target, ...sources) {
  for (source of sources.filter(Boolean)) {
    for (key of Object.keys(source)) {
      const descriptor = Object.getOwnPropertyDescriptor(source, key)
      if (descriptor.get) Object.defineProperty(target, key, descriptor)
      else target[key] = source[key]
    }
  }
}
```

So, it wouldn't be much work to change it, but it does sounds like something that would have to be considered more carefully.

ryanbrainard:
Ok, sounds like that's something to save for a later day, but the way this is now will make headers work the same way other things do, so merging it.

ryanbrainard:
Released in https://github.com/heroku/react-refetch/releases/tag/v1.0.1-1 (beta)

ryanbrainard:
Released in [v1.0.1](https://github.com/heroku/react-refetch/releases/tag/v1.0.1)

PR #184, opened 2017-10-06
Upgrade "warning" and "hoist-non-react-statics"

ipanasenko:


ryanbrainard:
Curious, what is the reason for upgrading these two? What issue is this fixing?

ipanasenko:
@ryanbrainard nothing, just two outdated dependencies, which caused multiple versions of same packages in my project, that's it

ryanbrainard:
Ok, I'll pull this in and put it out as a patch beta so folks and try it out

ryanbrainard:
Released in https://github.com/heroku/react-refetch/releases/tag/v1.0.3-0

PR #188, opened 2017-10-19
Update README.md

ipanasenko:
Add a note that `Array.prototype.find` polyfill is also needed (IE doesn't support it)


ryanbrainard:
Thanks!

Issue #154, opened 2016-12-04
Request.mode support


ryanbrainard:
Allow `Request.mode` to be set for no-cors (and other modes) support.

See https://github.com/heroku/react-refetch/pull/45

cc: @Adam-Burke

yhjor1212:
+1

alexcanessa:
+1

Any updates on this?

ryanbrainard:
Released in 1.0.4: https://github.com/heroku/react-refetch/releases/tag/v1.0.4

PR #190, opened 2017-11-20
Add support for Request.mode

ryanbrainard:
This has been requested a few times, so adding support for [`Request.mode`](https://developer.mozilla.org/en-US/docs/Web/API/Request/mode).

https://github.com/heroku/react-refetch/issues/154

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "5"
+  - "8"
 script:
   - npm run lint
   - npm test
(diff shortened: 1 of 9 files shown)


jsullivan:
This LGTM. I'm restarting Travis to see if those connection timeouts from the last build self-resolve.

jsullivan:
Hm, not sure what's up with Travis.

ryanbrainard:
@jsullivan turns out we were using a super ancient version of npm. I synced it up with what react-redux is using for versions.

jsullivan:
LGTM! Bring it on in. :)

PR #192, opened 2018-01-09
feat: add ability to provide function in identity requests (fixes #177)

### LINKED ISSUE #177: Allow value in identity requests to be a function
Currently in identity requests it's possible to pass a `Promise` into `value`, which is awesome, but if promise is returned from function, this function is called every time route changes, which leads to extra requests, even if user id didn't change, e.g. when navigating from `/user/:id/pulls` to `/user/:id/commits`

```js
const fetchUser = id => fetch(`/users/${id}`);

connect(props => ({
  user: {
    value: fetchUser(props.match.params.id),
    comparison: props.match.params.id
  }
}))(Profile);
```

This can be solved if `value` can be function:

```js
const fetchUser = id => fetch(`/users/${id}`);

connect(props => ({
  user: {
    value: () => fetchUser(props.match.params.id),
    comparison: props.match.params.id
  }
}))(Profile);
```

Function passed into `value` should only be invoked if comparison changed.
What do you think?

ipanasenko:
This PR fixes #177

It's now possible to pass a function to identity request's `value` prop:
```js
const fetchUser = id => fetch(`/users/${id}`);

connect(props => ({
  user: {
    value: () => fetchUser(props.match.params.id),
    comparison: props.match.params.id
  }
}))(Profile);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,7 +329,17 @@ connect(props => ({
 
 Note, this form of transformation is similar to what is possible on the `PromiseState` (i.e. `this.props.usersFetch.then(users => users.filter(u => u.verified))`); however, this has the advantage of only being called when `usersFetch` changes and keeps the logic out of the component.
 
-**Identity requests can also be provided a `Promise` or thenable.** In this case, the `PromiseState` will be `pending` until the `Promise` is resolved. This can be helpful for asynchronous, non-fetch operations (e.g. file i/o) that want to use a similar pattern as fetch operations.
+**Identity requests can also be provided a `Promise`/thenable or a `Function`.**
+In case of a `Promise`, the `PromiseState` will be `pending` until the `Promise` is resolved. This can be helpful for asynchronous, non-fetch operations (e.g. file i/o) that want to use a similar pattern as fetch operations.
+In case of a `Function`, value returned from it will be used instead, as in cases described above. `Function` will be only be called when `comparison` changed.
+
+```jsx
+connect(props => ({
+  usersFetch: {
+    value: () => someAPI.list('/users')
+  }
+}))(Users)
+```
 
 ## Accessing Headers & Metadata
 
@@ -564,7 +574,7 @@ This is an _advanced feature_. Use existing declarative functionality wherever p
 
 ## Unit Testing Connected Components
 
-For unit testing components connected, a non-default export of the unconnected component can be exposed to allow unit tests to inject their own `PromiseState`(s) as props. This allows for unit tests to test both success and error scenarios without having to deal with mocking HTTP, timing of responses, or other details about how the `PromiseState`(s) is fulfilled -- instead, they can just focus on asserting that the component itself renders the `PromiseState`(s) correctly in various scenarios. 
+For unit testing components connected, a non-default export of the unconnected component can be exposed to allow unit tests to inject their own `PromiseState`(s) as props. This allows for unit tests to test both success and error scenarios without having to deal with mocking HTTP, timing of responses, or other details about how the `PromiseState`(s) is fulfilled -- instead, they can just focus on asserting that the component itself renders the `PromiseState`(s) correctly in various scenarios.
 
 The recommended naming convention for the unconnected component is to prepend an underscore to the component name. For example, if there is a component called `Profile`, add a non-default export of `_Profile` before the default export with `connect`:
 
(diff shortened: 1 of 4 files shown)


ryanbrainard:
Thanks for working on this! Looks good at first glance, but will look into the tests and corner cases a little more later.

Since someone could have theoretically provided a function before, I guess this should be considered a breaking change..

Oh, can you update the API doc and readme for this change too?

ipanasenko:
@ryanbrainard updated docs

slorber:
@ipanasenko @ryanbrainard can we publish this as beta on npm? I'd be happy to test that implementation on a real project

ipanasenko:
@slorber I was very lazy, so I just built dist and copied it into my project's node_modules. You can fetch my branch and do the same ;)

ipanasenko:
@slorber or modify your package.json to look like this: `"react-refetch": "ipanasenko/react-refetch#feat/identity-function-dist",`

ryanbrainard:
@slorber @ipanasenko I just published this as a beta release: `react-refetch@2.0.0-0`

slorber:
thanks,

i've just tested and it seems to work fine

test/components/connect.spec.js
```diff
@@ -329,6 +329,135 @@ describe('React', () => {
...
+        const stub = TestUtils.findRenderedComponentWithType(container, Passthrough)
+        expect(stub.props.testFetch).toIncludeKeyValues({
+          fulfilled: true, pending: false, refreshing: false, reason: null, rejected: false, settled: true, value: 'foo', meta: { test: 'voodoo' }
+        })
+
+        done()
+      })
+    })
+
+    it('should invoke value() only if `comparison` changed', (done) => {
```
ryanbrainard: test name dupe. this looks like this was an accidential copy/paste of another test

README.md
```diff
@@ -329,7 +329,17 @@ connect(props => ({
...
 
-**Identity requests can also be provided a `Promise` or thenable.** In this case, the `PromiseState` will be `pending` until the `Promise` is resolved. This can be helpful for asynchronous, non-fetch operations (e.g. file i/o) that want to use a similar pattern as fetch operations.
+**Identity requests can also be provided a `Promise`/thenable or a `Function`.**
+In case of a `Promise`, the `PromiseState` will be `pending` until the `Promise` is resolved. This can be helpful for asynchronous, non-fetch operations (e.g. file i/o) that want to use a similar pattern as fetch operations.
+In case of a `Function`, value returned from it will be used instead, as in cases described above. `Function` will be only be called when `comparison` changed.
+
+```jsx
+connect(props => ({
+  usersFetch: {
+    value: () => someAPI.list('/users')
```
ryanbrainard: What about changing to fetching a single user with a `comparison` on the `props.userId` so it's more obvious when the function is called again?

test/components/connect.spec.js
```diff
@@ -329,6 +329,135 @@ describe('React', () => {
...
+      const renderSpy = expect.createSpy(() => ({}))
+      function render() {
+        renderSpy()
+        return <Passthrough/>
+      }
+
+      const valueSpy = expect.createSpy(() => ({}))
+      @connect(({ foo }) => ({
+        testFetch: {
+          value: valueSpy,
```
ryanbrainard: even though it won't actually change the behavior of the test, consider changing to `value: () => { valueSpy() }` to better test the function being created inside (like most apps would do), while still having the spy outside (so your test works)

ryanbrainard:
@ipanasenko I just pushed up some changes to https://github.com/heroku/react-refetch/compare/stage-2.0.0. I updated the docs and tests as I mentioned in my review. If you're happy with them, we could merge them as-is (or of course push any changes you don't like). Then we can do the warnings (if we want them... interested to hear your thoughts) in a separate PR.

slorber:
@ryanbrainard maybe it's worth issuing a warning in dev in case function change over time and comparison is not provided

ryanbrainard:
I finally dusted this off and added the `'Request object with functional 'value' must also declare 'comparison'` invariant to the `stage-2.0.0` branch. I'll merge that into master shortly.

ryanbrainard:
Published as beta v2.0.0-3

ryanbrainard:
Published in v2.0.0

PR #195, opened 2018-01-25
remove lodash to shrink bundle size

andriijas:
I added `react-refetch` to my app to playaround with it. Very nice! 

I noticed my bundle size increased a lot. Using `webpack-budle-analyzer` i found out that `react-refetch` included a good portion of lodash.

This PR replaces lodash with simpler methods to decrease the bundled size dramatically.

Before
before

After
after

PR #196, opened 2018-01-25
Fbjs

andriijas:
This PR builds upon #195  to decrease bundled size.
Given that react-refetch is used with React, we can use fbjs which React (invariant, shallowEqual) uses anyway so its almost certainly already exists in your bundle anyway

I added this as separate PR in case there is strong interest against this change. 

Before
before

After
after

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -148,8 +148,7 @@
     "asap": {
       "version": "2.0.6",
       "resolved": "https://registry.npmjs.org/asap/-/asap-2.0.6.tgz",
-      "integrity": "sha1-5QNHYR1+aQlDIIu9r+vLwvuGbUY=",
-      "dev": true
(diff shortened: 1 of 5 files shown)


ryanbrainard:
Looks great, but this warning on `fbjs` scares me a bit:

> Note: If you are consuming the code here and you are not also a Facebook project, be prepared for a bad time. APIs may appear or disappear and we may not follow semver strictly, though we will do our best to. This library is being published with our use cases in mind and is not necessarily meant to be consumed by the broader public. In order for us to move fast and ship projects like React and Relay, we've made the decision to not support everybody. We probably won't take your feature requests unless they align with our needs. There will be overlap in functionality here and in other open source projects.

Since we're using so few functions, it might be better to lift out what we need and remove the dep all together?

andriijas:
Then we can just disregard this pr and go for #195

Issue #214, opened 2018-08-07
Update Lifecycle Hooks for Future Async Rendering


ryanbrainard:
We will need to this:

https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#fetching-external-data

ideally before a "future 16.x release" and definitely before 17.0

villesau:
@ryanbrainard This prevents us from using e.g `getDerivedStateFromProps` with react-refetch. Using new lifecycle methods means that old lifecycle methods are not called:

```
vendor.js:504207 Warning: Unsafe legacy lifecycles will not be called for components using new component APIs.

Refetch.connect(MyComponent) uses getDerivedStateFromProps() but also contains the following legacy lifecycles:
  componentWillMount
  componentWillReceiveProps
  componentWillUpdate
```

ryanbrainard:
Merged in https://github.com/heroku/react-refetch/issues/231. It looks like https://github.com/heroku/react-refetch/pull/217 has been abandoned, but I am going to see if I can dust things off and move forward with this.

ryanbrainard:
Released in [3.0.0-1](https://github.com/heroku/react-refetch/releases/tag/v3.0.0-1 Beta)

PR #217, opened 2018-10-02
[WIP] Move away from deprecated lifecycle methods

tkvw:
This work in progress PR is intended to discuss and update refetch to move away from deprecated `componentWillMount` and `componentWIllReceiveProps` methods to `componentDidMount` and `componentDidUpdate`. Also context support is removed. 

Motivation: 
* To be inline with latest of React
* Support SSR (componentWillMount is called on `renderToString`)

Tasks: 
- [x] Update component
- [ ] Discuss PR 
- [ ] Update tests

### DIFF
--- a/src/components/connect.js
+++ b/src/components/connect.js
@@ -67,8 +67,10 @@ const omitChildren = function omitChildren(obj) {
 function connect(mapPropsToRequestsToProps, defaults, options) {
   const finalMapPropsToRequestsToProps = mapPropsToRequestsToProps || defaultMapPropsToRequestsToProps
   const dependsOnProps = finalMapPropsToRequestsToProps.length >= 1
-  const dependsOnContext = finalMapPropsToRequestsToProps.length == 2
-
+  
+  warning(finalMapPropsToRequestsToProps.length < 2, 'Passing context to `mapPropsToRequestsToProps` is no longer supported.')
+  
+  
   let topFetch
   let topRequest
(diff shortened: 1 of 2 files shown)


ryanbrainard:
@tkvw Nice to see someone working on this :)

https://github.com/heroku/react-refetch/issues/214

tkvw:
I decided to go with my own custom redux based implementation; this is on my to-do list, but don't expect it to be finished soon.

src/components/connect.js
```diff
@@ -413,21 +414,18 @@ function connect(mapPropsToRequestsToProps, defaults, options) {
...
+          previous.apply(this, arguments)
           return
         }
 
         // We are hot reloading!
         this.version = version
         this.clearAllRefreshTimeouts()
         this.refetchDataFromProps()
-      }
+      })(RefetchConnect.prototype.componentDidUpdate)
```
ryanbrainard: @tkvw I know you've moved on from this PR (thanks for getting it started BTW!), but I've been working on reviving it to hopefully fix this once and for all. I have a question about the change to this section. Do you happen to remember why you needed to call `previous.apply(this, arguments)` in here?  It seems like if `this.version === version`, this should not be needed, but I could very easily be missing something.

ryanbrainard:
Released in [3.0.0-1 Beta](https://github.com/heroku/react-refetch/releases/tag/v3.0.0-1)

PR #220, opened 2018-11-27
Publish Types

ryanbrainard:
#219 introduced types, but as mentioned in https://github.com/heroku/react-refetch/pull/219#issuecomment-437950905, they aren't actually published anywhere.

This makes the following changes (a little different than discussed in https://github.com/heroku/react-refetch/pull/219#issuecomment-439274669, but I think this way is cleaner):

- Move `index.d.ts` into `src`, next to `index.js`
- At build time, move `index.d.ts` to `lib`, next to the transpiled `index.js`
- Register `./lib/index.d.ts` as main type file in `package.json`

To help test this out, I have published this as beta version `2.0.2-0`. @lmontoute or @hburrows, can you test this out on your projects?

cc:  @ipanasenko


lmontoute:
References fine for me in VSCode. Tested by deleting the types we're using, right clicking an import and going to declaration, which takes me to `./lib/index.d.ts`.

We don't use every library feature exhaustively but I imported everything and there were no compiler errors across our whole project.

I think this is good to go 🙂

PR #223, opened 2019-01-03
Update body type

### LINKED ISSUE #222: [Typescript] Type of "body" attribute must be string? and not object? on the Mapping interface
As describe in documentation when implementing a POST and providing the body attribute you should do something like:

```
connect(() => ({
        login: (signInParams: ISignInParams) => ({
            loginResponse: {
                url: `${host}/session`,
                method: 'POST',
                body: JSON.stringify(signInParams)
            }
        })
    }))(newComponent)
```

However, this snippets doesn't compile with Typescript since the body is expected to be an **object?**.

Therefore it raise the following compilation error:

> Type '(signInParams: ISignInParams) => { loginResponse: { url: string; method: string; body: string; }; }' is not assignable to type '(signIn: ISignInParams) => PropsMap<SignInProps>'.
>   Type '{ loginResponse: { url: string; method: string; body: string; }; }' is not assignable to type 'PropsMap<SignInProps>'.
>     Types of property 'loginResponse' are incompatible.
>       Type '{ url: string; method: string; body: string; }' is not assignable to type 'string | Mapping<SignInProps, {}> | undefined'.
>         Type '{ url: string; method: string; body: string; }' is not assignable to type 'Mapping<SignInProps, {}>'.
>           Types of property 'body' are incompatible.
>             Type 'string' is not assignable to type 'object | undefined'.

raphaelpra:
Fix #222

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -105,7 +105,7 @@ export interface Mapping<TProps, TValue> {
   method?: string;
   headers?: { [key: string]: string | (() => string) };
   credentials?: "omit" | "same-origin" | "include";
-  body?: object;
+  body?: string;
   redirect?: "follow" | "error" | "manual";
   mode?: "cors" | "no-cors" | "same-origin" | "navigate";
   refreshInterval?: number;


ryanbrainard:
This is definitely better than `object`, but the `Request` [docs](https://developer.mozilla.org/en-US/docs/Web/API/Request) say:

> The body type can only be a Blob, BufferSource, FormData, URLSearchParams, USVString or ReadableStream type, so for adding a JSON object to the payload you need to stringify that object.

so wondering if we want to list these or just use `Any`... Thoughts?

ryanbrainard:
I'm going to merge this for now because its better than what we have now and covers the general case. We can always go back and add more types if needed.

ryanbrainard:
Released in v2.0.3. Thanks @raphaelpra !!! 🎉

raphaelpra:
Thanks for the merge !

I agree with remark regarding the `Request`. However I do believe that `any` type is misleading. Since it is a common mistake to put an object instead of string in body of a request (I have done so many times). I appreciate having a type validation in this case.

I will have a look on how we can allow other types (Blob, FormData, ...). Ideally, I think we should infer the type from Fetch library used. Not sure how to do it, but I'll have a look. Thoughts ?

raphaelpra:
Maybe something like that -> https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node-fetch/index.d.ts#L168 without `NodeJS.ReadableStream` (backend logic) and `URLSearchParams ` (that would require the new dependency 'url')

ryanbrainard:
That sounds reasonable, although it strange that it doesn't include other things like `Blob` and `FormData`. 

[Here](https://github.com/ef-carbon/fetch) is another I found. They also have `BufferSource`, which is defined as `ArrayBuffer | Int8Array | Uint8Array | Uint8ClampedArray | Int16Array | Uint16Array | Int32Array | Uint32Array | Float32Array | Float64Array`. This one seems a little more complete.

PR #240, opened 2019-11-18
Improve type definition

ykikura-indeed:
This change improve TS type definition bundled in react-refetch.

## Background

Follow code will throw TypeError, since `userFetch.value` is undefined until data is loaded.

```tsx
render() {
  return <span>{ this.props.userFetch.value.name }</span>
}
```

It's better to detect this kind of bug by type-checking

## Solution

This change split `PromiseState` into 3 types, `PendingPromiseState`, `FulfilledPromiseState`, and `RejectedPromiseState`.
The property `value` exists only in `FulfilledPromiseState`. Therefore, before checking whether `promiseState` is `FulfilledPromiseState` instance or not, accessinig `promiseState.value` become compile-error. 

```tsx
render() {
  if (!this.props.userFetch.fulfilled) {
    return <span>Loading...</span>
  }

  // Here, TS compiler knows "userFetch" is "FulfilledPromiseState" instance, because "uerFetch.fulfilled" is "true"
  return <span>{ this.props.userFetch.value.name }</span>
}
```

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -21,14 +21,12 @@ export interface PromiseStateStatic {
   race<T = {}>(iterable: Iterable<PromiseState<any>>): PromiseState<T>;
 }
 
-export interface PromiseState<T = {}> {
+interface PromiseStateBase<T = {}> {
   readonly pending: boolean;
   readonly refreshing: boolean;
   readonly fulfilled: boolean;
   readonly rejected: boolean;
   readonly settled: boolean;
-  readonly value: T;
-  readonly reason: any;
   readonly meta: any;
   then: <TFulfilled = T, TRejected = T>(
     onFulfilled?: (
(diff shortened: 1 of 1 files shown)


src/index.d.ts
```diff
@@ -44,6 +42,28 @@ export interface PromiseState<T = {}> {
   ) => PromiseStateLike<T> | PromiseStateLike<TRejected>;
 }
 
+export interface PendingPromiseState<T = {}> extends PromiseStateBase {
+  readonly pending: true;
+  readonly fulfilled: false;
+  readonly rejected: false;
+}
+
+export interface FulfilledPromiseState<T = {}> extends PromiseStateBase {
```
ryanbrainard: would it make sense to change Line 18 to:

```tsx
resolve<T = {}>(value?: PromiseStateLike<T>, meta?: any): FulfilledPromiseState<T>;
```

and likewise for the other static methods? Otherwise, those will all be `PendingPromiseState<T>`, right?

I'm wondering how this is going to impact the static `all` and `race` methods. I'm kind of wondering if we need to still have a generic type for those...

ykikura-indeed: Yes, that's much better! I also update other static methods excepts `all` and `race`. I think it's not so much helpful to introduce generic interfaces for these 2 methods.

ykikura-indeed:
I think this technique is commonly used in TypeScript, to maximize the benefit of using TypeScript.

> other places generic PromiseStates are expected

There is no problem, I think. We can assign each specific promise state type into generic PromiseState variables.

```ts
if (p1.pending) {
  // Here, p1 is `PendingPromiseState<T>`.

  // All follow statements are valid
  const p2: PromiseState<T> = pendingPromiseState;
  func(p1);
}

function func(p: PromiseState<T>) {
}
```

ryanbrainard:
I've tried this out and it seems to work fine. I'll try to get this out in a bigger change to fix some other TypeScript issues. Thanks, again!

PR #241, opened 2019-11-18
Fix value's type

ykikura-indeed:
Simple type definition fix to use promise factory method in "value" parameter.


ryanbrainard:
Oops, sorry I didn't actually merge this before!

Issue #230, opened 2019-07-23
Typescript types seem not to work properly


yannick-cw:
Hey, I tried to integrate react-refetch to our stack and could not get it to work with typescript properly. To me it looks like there are some problems with the type definitions, but maybe I missed something on usage side.

Here is a short example

```js
import React from 'react';
import { connect, PromiseState } from 'react-refetch';

interface Props {
  exampleFetch: PromiseState<any>;
  exampleUpdate: (term: string) => void;
  exampleUpdateDone?: PromiseState<string>;
}

export class SmallExampleComp extends React.PureComponent<Props> {
  public render() {
    return <div>ok</div>;
  }
}

const ConnectedComp = connect<Props>(_props => ({
  exampleFetch: 'url',
  exampleUpdate: (_term: string) => ({
    exampleUpdateDone: {
      url: 'url',
    },
  }),
}))(SmallExampleComp);

export const Usage = <ConnectedComp />;
```

This fails to compile at two places
First at usage side it seems to still require the properties, that are actually added in the connect:
```
Type '{}' is missing the following properties from type 'Readonly<Props>': exampleFetch, exampleUpdate

26 export const Usage = <ConnectedComp />;
```

Second the `?` optional on the exampleUpdateDone does not work with the connect, but is required, as the `exampleUpdateDone` is optional.
```
    Types of property 'exampleUpdateDone' are incompatible.
      Type '{ url: string; }' is not assignable to type 'undefined'.

19   exampleUpdate: (_term: string) => ({
```
If I am missing something please give me a hint, otherwise I'd like to help fix these type definitions.

yannick-cw:
Okay I have fixed the types for us locally with this version:

```javascript
import { ComponentClass, FunctionComponent } from 'react';
import { connect as badConnect, Mapping, WithRefetch } from 'react-refetch';

type FixedConnect = <FetchProps, CompProps>(
  map: MapPropsToRequestsToProps<FetchProps, CompProps>,
) => (
  component: ComponentClass<CompProps> | FunctionComponent<CompProps>,
) => ComponentClass<Omit<CompProps, keyof FetchProps>> & WithRefetch<Omit<CompProps, keyof FetchProps>>;

type MapPropsToRequestsToProps<FProps, CProps> = (props: Omit<CProps, keyof FProps>, context: any) => PropsMap<FProps>;

type PropsMap<FProps> = {
[...]

PR #242, opened 2019-11-27
Typescript: Omit Fetch Props from Component Props

ryanbrainard:
This pulls in @yannick-cw's change from https://github.com/heroku/react-refetch/issues/230.

I first committed a small example in ab96a2a to repro the problem and then applied the fix in d5ae588. 

@yannick-cw, @ddanielbee: Can you take a look at the `UserWidget` example and let me know if this is how you were using it, particularly how I defined `FetchProps` and `CompProps` and how I passed them in as `connect<FetchProps, CompProps>`. It compiles and all seems to work, but just wanted to make sure I got it correct as an example.

Fixes https://github.com/heroku/react-refetch/issues/230


src/index.d.ts
```diff
@@ -88,30 +88,24 @@ export interface ConnectOptions {
...
+) => PropsMap<FProps>;
 
 // String or PromiseState
-type PromiseStateMapping<
-  TProps,
-  TProp extends keyof TProps
-> = TProps[TProp] extends PromiseState<infer TValue>
-  ? string | Mapping<TProps, TValue>
-  : never;
+type PromiseStateMapping<FProps> = string | Mapping<FProps, any>;
```
ryanbrainard: should this have a ` | never` on the end?

ryanbrainard:
I was just writing up readme docs how to use the TypeScript types, and I realized the whole `Omit` behavior is unintuitive and pretty hard to explain because the props exposed to callers are subtractive. To improve this, I tried flipping things around. Instead of `connect` magically omitting the fetch props, the component defines `OuterProps` (raw data) and `InnerProps extends OuterProps` (raw data + fetches). This avoids the use of `Omit` and uses explicit types to make it easier both for the component and caller what is expected to be passed vs what the component has access to internally. I added an explanation in the readme as well the shows an example using this pattern. Thoughts on this change?

ryanbrainard:
I just released this branch as beta `react-refetch@4.0.0-0` in case anyone would like to try it out before it is merged. This is a major version because of the breaking changes for TypeScript users (no changes to the lib itself).

ryanbrainard:
Review was done in https://github.com/heroku/react-refetch/issues/230#issuecomment-560286529

PR #244, opened 2020-01-19
TypeScript: Improve type definition to use optional chaining

Kiikurage:
[Optional Chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) is the syntax in upcoming ECMAScript and introduced in TypeScript v3.7.

In #240, I(Sorry, account is different but it's me) changed react-refetch's type definition so that unsafe access can be detected by TS compiler.

```ts
render() {
  if (!this.props.userFetch.fulfilled) {
    return <span>Loading...</span>
  }

  // Here, TS compiler knows "userFetch" is "FulfilledPromiseState" instance, because "uerFetch.fulfilled" is "true"
  return <span>{ this.props.userFetch.value.name }</span>
}
```

However, sometimes people want to simplify this by using optional chaining

```ts
render() {
  return <span>{ this.props.userFetch.value?.name || 'Loading...' }</span>
}
```

This PR improve type definition so that all `PromiseState` type have value field to use optional chaining for generic `PromiseState` type.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -47,6 +47,8 @@ export interface PendingPromiseState<T = {}> extends PromiseStateBase {
   readonly pending: true;
   readonly fulfilled: false;
   readonly rejected: false;
+  readonly value: undefined;
+  readonly reason: undefined;
 }
 
 export interface FulfilledPromiseState<T = {}> extends PromiseStateBase {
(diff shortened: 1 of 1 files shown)


ryanbrainard:
Thanks, this makes sense.

ryanbrainard:
Released in https://github.com/heroku/react-refetch/releases/tag/v4.0.1-0

Going to let that bake a little bit and then we can push out v4 as GA

PR #272, opened 2024-06-04
feat: prepare for packaging

mble-sfdc:
This commit updates the name of the package and some of its metadata in preparation for publication under the `heroku` npm org.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "react-refetch",
+  "name": "@heroku/react-refetch",
   "version": "5.0.0-0",
   "description": "A simple, declarative, and composable way to fetch data for React components.",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -15,7 +15,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/heroku/react-refetch.git"
+    "url": "git+https://github.com/heroku/react-refetch.git"
```
ryanbrainard: What is this change for?

mble-sfdc: This was automatic as part of the pre-prep. See https://docs.npmjs.com/cli/v10/configuring-npm/package-json#git-urls-as-dependencies. Unclear as this shouldn't be valid for the `repository` object. I'll remove.

